PR #9, opened 2019-04-15
fix(sudo-tag): remove sudo tag from travis

ehdsouza:
fixes https://github.com/IBM/python-sdk-core/issues/8


ehdsouza:
This is already xenial

PR #10, opened 2019-04-19
feat: use iam.cloud.ibm.com in IAM url; allow IAM client_id/secret to…

padamstx:
… be configured

Fixes: https://github.ibm.com/arf/planning-sdk-squad/issues/851

### DIFF
--- a/ibm_cloud_sdk_core/iam_token_manager.py
+++ b/ibm_cloud_sdk_core/iam_token_manager.py
@@ -19,17 +19,18 @@ import time
 from .api_exception import ApiException
 
 class IAMTokenManager(object):
-    DEFAULT_IAM_URL = 'https://iam.bluemix.net/identity/token'
+    DEFAULT_IAM_URL = 'https://iam.cloud.ibm.com/identity/token'
     CONTENT_TYPE = 'application/x-www-form-urlencoded'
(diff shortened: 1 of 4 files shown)


ibm_cloud_sdk_core/iam_token_manager.py
```diff
@@ -39,9 +40,12 @@ def __init__(self, iam_apikey=None, iam_access_token=None, iam_url=None):
         }
 
     def request(self, method, url, headers=None, params=None, data=None, **kwargs):
+        auth_tuple = ('bx', 'bx')
```
ehdsouza: Why is the default `bx`? `None` would be better if it doesnt really have a default

padamstx: Our current default Authorization header is the result of doing a base64 encoding of 'bx:bx'.  In other words, the default client_id and secret of 'bx' and 'bx' are already being used.   It's just that we were simply hardcoding the resulting value of the Authorization header ("Basic Yng6Yng=").  Now that we've exposed client_id and secret parameters so they can be explicitly configured, it makes sense to supply the Authorization header in a slightly different way (i.e. let the requests package compute the actual header value for us by instead just supplying the "auth" parameter as a tuple (<client_id>, <secret>).

PR #14, opened 2019-05-08
fix(detailed response): Update _to_dict function to support JSON array results

### LINKED ISSUE #13: DetailedResponse's _to_dict function does not properly handle JSON arrays
If the JSON response is an array, the `_to_dict` will return `HTTP Response`:

Code snippet from `DetailedResponse`:
```
    def _to_dict(self):
        _dict = {}
        if hasattr(self, 'result') and self.result is not None:
            _dict['result'] = self.result if isinstance(self.result, dict) else 'HTTP response'
        ...
``` 
What I'd expect:
The response object returns an array of items

A quick fix would be to check if the `result` is an instance of a `list`:
```
        if hasattr(self, 'result') and self.result is not None:
            _dict['result'] = self.result if isinstance(self.result, dict) or isinstance(self.result, list)  else 'HTTP response'
```

emlaver:
fixes #13

### DIFF
--- a/ibm_cloud_sdk_core/base_service.py
+++ b/ibm_cloud_sdk_core/base_service.py
@@ -198,7 +198,7 @@ class BaseService(object):
         self.password = password
         self.jar = CookieJar()
 
-    def set_token_manager(self, iam_apikey=None, iam_access_token=None, iam_url=None, 
+    def set_token_manager(self, iam_apikey=None, iam_access_token=None, iam_url=None,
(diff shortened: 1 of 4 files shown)


ehdsouza:
@emlaver this PR looks good! Thanks for your contribution 👍 

I am wondering if we even need to support `str`.

emlaver:
@ehdsouza Thanks for reviewing!

PR #21, opened 2019-08-16
Auth

ehdsouza:
This PR supports the new authentication mechanism, an example:
```py
from ibm_cloud_sdk_core.authenticators import BasicAuthenticator

assistant = AssistantV1(version='2018-07-10', 
                                        authenticator=BasicAuthenticator('<username>', '<password>'))
```


dpopp07:
Haven't looked at the code yet, but have a couple questions about the release strategy:

I see you're merging this into `master`, so will that trigger a release? Isn't this a breaking change? If this is going into the major release candidate for the core, perhaps it should go into a `next` branch or something like that.

That is, unless this is the _only_ change you want to go into the next major release. If that's the case, we can just do the major release with this PR like @padamstx did with Java.

Either way, shouldn't one of the commit messages have the `BREAKING CHANGE` tag on it?

ibm_cloud_sdk_core/authenticators/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from .authenticator import Authenticator
+from .basic_authenticator import BasicAuthenticator
+from .bearer_authenticator import BearerAuthenticator
```
dpopp07: We are calling this `BearerTokenAuthenticator`

ibm_cloud_sdk_core/authenticators/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+from .cp4d_authenticator import CP4DAuthenticator
```
dpopp07: We are calling this `CloudPakForDataAuthenticator`

ibm_cloud_sdk_core/authenticators/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+from .iam_authenticator import IAMAuthenticator
+from .no_auth_authenticator import NoAuthAuthenticator
```
dpopp07: We are calling this `NoauthAuthenticator`

ibm_cloud_sdk_core/authenticators/basic_authenticator.py
```diff
@@ -0,0 +1,78 @@
...
+        self.password = password
+        self.validate()
+
+    def set_username_and_password(self, username, password):
+        """
+        Sets the username and password
+        """
+        self.username = username
+        self.password = password
+        self.validate()
```
dpopp07: `username` and `password` are not settable according to the design document

ibm_cloud_sdk_core/authenticators/bearer_authenticator.py
```diff
@@ -0,0 +1,59 @@
...
+        """
+        Performs validation on input params
+        """
+        if self.bearer_token is None:
+            raise ValueError('The bearer token shouldn\'t be None.')
+
+        if has_bad_first_or_last_char(self.bearer_token):
+            raise ValueError(
+                'The bearer token shouldn\'t start or end with curly brackets or quotes. '
+                'Please remove any surrounding {, }, or \" characters.')
```
dpopp07: This wasn't explicit, but I don't think we need to check the first and last character on a beaker token. I don't know that they would ever have those characters, but because they are auto-generated I feel weird about checking for something that could be there and be correct.

I could be wrong about that though

ibm_cloud_sdk_core/authenticators/cp4d_authenticator.py
```diff
@@ -0,0 +1,112 @@
...
+        """
+        self.token_manager.set_password(password)
+        self.validate()
+
+    def set_url(self, url):
+        """
+        Sets the url
+        """
+        self.token_manager.set_url(url)
+        self.validate()
```
dpopp07: `username`, `password`, and `url` are not settable

ibm_cloud_sdk_core/authenticators/cp4d_authenticator.py
```diff
@@ -0,0 +1,112 @@
...
+        self.token_manager = CP4DTokenManager(
+            username, password, url, disable_ssl_verification, headers, proxies)
+        self.validate()
+
+    def validate(self):
+        """
+        Performs validation on input params
+        """
+        if self.token_manager.username is None or self.token_manager.password is None:
+            raise ValueError('The username and password shouldn\'t be None.')
```
dpopp07: `url` is also a required parameter

ibm_cloud_sdk_core/authenticators/iam_authenticator.py
```diff
@@ -0,0 +1,120 @@
...
+        Set the IAM api key
+        """
+        self.token_manager.set_apikey(apikey)
+        self.validate()
+
+    def set_url(self, url):
+        """
+        Set the IAM url
+        """
+        self.token_manager.set_url(url)
```
dpopp07: `apikey` and `url` are not settable

ibm_cloud_sdk_core/base_service.py
```diff
@@ -23,134 +23,72 @@
...
+                    vcap_service_credentials, dict):
+                if vcap_service_credentials.get('username') and vcap_service_credentials.get('password'): # cf
+                    vcap_service_credentials['auth_type'] = 'basic'
+                elif vcap_service_credentials.get('apikey'): # rc
+                    vcap_service_credentials['auth_type'] = 'iam'
+                self._set_authenticator_properties(vcap_service_credentials)
+                self._set_service_properties(vcap_service_credentials)
+
+        if not self.authenticator:
+            self.authenticator = NoAuthAuthenticator()
```
dpopp07: We decided that an `authenticator` would be `required` in the base service. If one is not set, rather than use a default, we throw an error. In the `generated` service for Watson, we will specify the default of using the environment to create an authenticator. I think you need a separate module for this, since you user should also be able to get an authenticator like this without relying on the default.

ehdsouza:
@dpopp07 I changed the base to `next` as I would be making more minor changes like updating the request library.

ibm_cloud_sdk_core/authenticators/authenticator.py
```diff
@@ -0,0 +1,46 @@
...
+        Return true if basic authentication
+        """
+        pass
+
+    @abstractmethod
+    def _is_bearer_authentication(self):
+        """
+        Return true if bearer authentication
+        """
+        pass
```
dpopp07: What are these methods `_is_bearer_authentication` `_is_basic_authentication` used for? I don't know that we necessarily need them

ibm_cloud_sdk_core/base_service.py
```diff
@@ -365,52 +267,62 @@ def request(self, method, url, accept_json=False, headers=None,
...
             error_message = None
             if response.status_code == 401:
                 error_message = 'Unauthorized: Access is denied due to ' \
                                 'invalid credentials'
-            raise ApiException(response.status_code, error_message, http_response=response)
+            raise ApiException(
+                response.status_code, error_message, http_response=response)
 
     @staticmethod
     def _convert_model(val, classname=None):
```
dpopp07: Love to see all the code removed in this file!

padamstx:
@ehdsouza I haven't had a chance to look at the changes in detail yet, but could you add python code examples to the [design wiki](https://github.ibm.com/CloudEngineering/openapi-sdkgen/wiki/Core-SDK-Authentication-Design).
That will likely help us a bit during the review, plus it would be good to have examples for all the languages on the wiki :)

ibm_cloud_sdk_core/authenticators/iam_authenticator.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def set_authorization_info(self, client_id, client_secret):
```
dpopp07: This should be called `setClientIdAndSecret`. We left this out of the design document by mistake but I've added it there now!

ibm_cloud_sdk_core/authenticators/bearer_authenticator.py
mediumTaj: I think you can also get the token from the cp4d console - maybe we should be checking for bad characters here? I don't remember what getting the credentials looks like in the console.

ibm_cloud_sdk_core/authenticators/basic_authenticator.py
```diff
@@ -0,0 +1,78 @@
...
+                self.username) or has_bad_first_or_last_char(self.password):
+            raise ValueError(
+                'The username and password shouldn\'t start or end with curly brackets or quotes. '
+                'Please remove any surrounding {, }, or \" characters.')
+
+    def authenticate(self):
+        """
+        Returns the username and password tuple
+        """
+        return (self.username, self.password)
```
mediumTaj: In the other authenticators the authenticate function returns a string to use for the Authorization header

````python
return 'Bearer {0}'.format(self.bearer_token)
````

Should this return the base 64 encoded u:p? Maybe we can get rid of `_is_bearer_authentication` and `_is_basic_authentication`?

Comment (dpopp07) [reply ibm_cloud_sdk_core/authenticators/basic_authenticator.py]: > In the other authenticators the authenticate function returns a string to use for the Authorization header

Actually, in Node, I'm passing the request object to the authenticate method and adding the header to it by reference. Maybe that's something that I'll be asked to change but how the authenticate method works could potentially be different from language to language, since it's not user facing. Could be worth discussing.

File: ibm_cloud_sdk_core/authenticators/bearer_authenticator.py (hunk 5 above)
Comment (dpopp07) [reply ibm_cloud_sdk_core/authenticators/bearer_authenticator.py]: Yeah maybe we should look at examples but it's still a generated string. Probably will never be a bracket or quotation at the beginning or end so this is probably not a bad thing but since we can't control it, it might be risky. And I don't know that it adds a lot of value for the risk

File: ibm_cloud_sdk_core/authenticators/basic_authenticator.py (hunk 13 above)
Comment (ehdsouza) [reply ibm_cloud_sdk_core/authenticators/basic_authenticator.py]: So in python `request` library, the basic auth can be passed directly without encoding using the auth=(username, password) param. And for a bearer token, we have to pass it as a special `Authorization` header.

File: ibm_cloud_sdk_core/authenticators/bearer_authenticator.py (hunk 5 above)
Comment (lpatino10) [reply ibm_cloud_sdk_core/authenticators/bearer_authenticator.py]: Yeah we shouldn't have to. Here's a screenshot of the screen where you get the token and URL in CP4D:

Screen Shot 2019-08-19 at 10 32 19 AM

It's basically the same as in public where you just hit a copy button to grab the right values.

File: ibm_cloud_sdk_core/authenticators/basic_authenticator.py (hunk 13 above)
Comment (mediumTaj) [reply ibm_cloud_sdk_core/authenticators/basic_authenticator.py]: @dpopp07 yeah thats how .NET works as well, by other authenticators i mean the python bearerToken, cp4d and iam authenticators all return strings to add to the header.

Comment (dpopp07) [reply ibm_cloud_sdk_core/authenticators/basic_authenticator.py]: That makes a lot of sense! My bad on the misunderstanding

File: ibm_cloud_sdk_core/authenticators/basic_authenticator.py [hunk 14]
```diff
@@ -0,0 +1,78 @@
...
+
+    def set_username_and_password(self, username, password):
+        """
+        Sets the username and password
+        """
+        self.username = username
+        self.password = password
+        self.validate()
+
+    def _is_basic_authentication(self):
```
padamstx: I think we can get rid of these "_is_xxx" functions.

ibm_cloud_sdk_core/authenticators/bearer_authenticator.py
padamstx: My vote is to avoid perform those "HasBadCharacters..." checks.  Those seem to be a remnant of a by-gone era :)  Hopefully...

ibm_cloud_sdk_core/authenticators/bearer_authenticator.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def authenticate(self):
+        """
+        Returns the bearer token
+        """
+        return 'Bearer {0}'.format(self.bearer_token)
```
padamstx: Again... authenticate shouldn't return anything... it should accept a Request object (whatever that means for python) and update that object as needed to perform the authentication.

ibm_cloud_sdk_core/authenticators/iam_authenticator.py
```diff
@@ -0,0 +1,120 @@
...
+        """
+        :attr str apikey: The apikey
+        :attr str url: The url for authentication
+        :attr str client_id: The client id for rate limiting
+        :attr str client_secret: The client secret for rate limiting
+        :attr bool disable_ssl_verification: enables/ disabled ssl verification
+        :attr dict headers: user-defined headers
+        :attr dict proxies: user-defined proxies
+        """
+        self.token_manager = IAMTokenManager(
```
padamstx: I think you should be able to collapse the IAMTokenManager and IamAuthenticator classes together, no?  They should be one and the same I think.

ibm_cloud_sdk_core/authenticators/no_auth_authenticator.py
```diff
@@ -0,0 +1,32 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from .authenticator import Authenticator
+
+class NoAuthAuthenticator(Authenticator):
```
padamstx: This should be called NoauthAuthenticator.

ibm_cloud_sdk_core/authenticators/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from .authenticator import Authenticator
+from .basic_authenticator import BasicAuthenticator
+from .bearer_authenticator import BearerAuthenticator
+from .cp4d_authenticator import CP4DAuthenticator
+from .iam_authenticator import IAMAuthenticator
```
padamstx: Should be called IamAuthenticator

ibm_cloud_sdk_core/authenticators/iam_authenticator.py
```diff
@@ -0,0 +1,120 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from .authenticator import Authenticator
+from ..iam_token_manager import IAMTokenManager
+from ..utils import has_bad_first_or_last_char
+
+
+class IAMAuthenticator(Authenticator):
```
padamstx: IamAuthenticator

ibm_cloud_sdk_core/base_service.py
```diff
@@ -23,134 +23,72 @@
...
-    IAM_ACCESS_TOKEN = 'iam_access_token'
-    ICP4D_ACCESS_TOKEN = 'icp4d_access_token'
-    URL = 'url'
-    USERNAME = 'username'
-    PASSWORD = 'password'
-    IAM_APIKEY = 'iam_apikey'
-    IAM_URL = 'iam_url'
-    ICP4D_URL = 'icp4d_url'
-    APIKEY_DEPRECATION_MESSAGE = 'Authenticating with apikey is deprecated. Move to using Identity and Access Management (IAM) authentication.'
     DEFAULT_CREDENTIALS_FILE_NAME = 'ibm-credentials.env'
```
padamstx: We should no longer have any references to credential file-related things in BaseService.
The BaseService class will need to retrieve any properties related to the service (based on service name) so that it can obtain the service-level URL and disableSSL properties, but hopefully BaseService doesn't need to concern itself with any of those details other than just use a utility class to retrieve the properties.

ibm_cloud_sdk_core/base_service.py
```diff
@@ -23,134 +23,72 @@
...
+        if self.authenticator:
+            if not isinstance(self.authenticator, Authenticator):
+                raise ValueError(
+                    'authenticator should be of type Authenticator')
 
         # 2. Credentials from credential file
-        if display_name and not self.username and not self.token_manager:
+        if display_name and not self.authenticator:
             service_name = display_name.replace(' ', '_').lower()
             self._load_from_credential_file(service_name)
```
padamstx: Nope, we don't do this from BaseService any more.

ibm_cloud_sdk_core/authenticators/iam_authenticator.py
ehdsouza: Prefer adding it as a separate token manager class

README.md
```diff
@@ -21,17 +21,52 @@ easy_install --upgrade ibm-cloud-sdk-core
 ```
 
 ## Authentication Types
-There are several flavors of authentication supported in this package. To specify the intended authentication pattern to use, the user can pass in the parameter `authentication_type`. This parameter is optional, but it may become required in a future major release. The options for this parameter are `basic`, `iam`, and `icp4d`.
+There are several flavors of authentication supported in this package. To specify the intended authentication pattern to use, simply instantiate a `Authenticator` of your choice.
```
dpopp07: Tiny wording suggestion
```suggestion
There are several flavors of authentication supported in this package. To specify the intended authentication pattern to use, simply instantiate the `Authenticator` of your choice.
```

ibm_cloud_sdk_core/authenticators/iam_authenticator.py
```diff
@@ -0,0 +1,102 @@
...
+        :attr str client_id: The client id for rate limiting
+        :attr str client_secret: The client secret for rate limiting
+        :attr bool disable_ssl_verification: enables/ disabled ssl verification
+        :attr dict headers: user-defined headers
+        :attr dict proxies: user-defined proxies
+        """
+        self.token_manager = IAMTokenManager(
+            apikey, url, client_id, client_secret, disable_ssl_verification,
+            headers, proxies)
+        self.validate()
```
dpopp07: I wonder if you should validate before you create the token manager, so that you're not passing a null value for apikey if none is given. Probably doesn't make a big difference though

ibm_cloud_sdk_core/authenticators/iam_authenticator.py
```diff
@@ -0,0 +1,102 @@
...
+        headers['Authorization'] = 'Bearer {0}'.format(bearer_token)
+
+    def set_client_id_and_secret(self, client_id, client_secret):
+        """
+        Set the IAM authorization information.
+        This consists of the client_id and secret.
+        These values are used to form the basic authorization header that
+        is used when interacting with the IAM token server.
+        If these values are not supplied, then a default Authorization header
+        is used.
```
dpopp07: > If these values are not supplied, then a default Authorization header is used.

We are actually not using a default header anymore so this description should change. We're just not sending the auth header (it's not required).

ibm_cloud_sdk_core/base_service.py
```diff
@@ -22,284 +22,100 @@
...
-from .icp4d_token_manager import ICP4DTokenManager
+from .utils import has_bad_first_or_last_char, remove_null_values, cleanup_values, read_from_external_sources
 from .detailed_response import DetailedResponse
 from .api_exception import ApiException
-
-try:
-    from http.cookiejar import CookieJar  # Python 3
-except ImportError:
-    from cookielib import CookieJar  # Python 2
+from .authenticators import Authenticator, BasicAuthenticator, BearerTokenAuthenticator, CloudPakForDataAuthenticator, IamAuthenticator, NoauthAuthenticator
```
dpopp07: Why are you importing all of the authenticators here? I'm not sure you need them

ibm_cloud_sdk_core/base_service.py
```diff
@@ -22,284 +22,100 @@
...
-        if vcap_services is not None:
-            services = json_import.loads(vcap_services)
-            if service_name in services:
-                return services[service_name][0]['credentials']
-        else:
-            return None
+        if self.authenticator:
+            if not isinstance(self.authenticator, Authenticator):
+                raise ValueError(
+                    'authenticator should be of type Authenticator')
```
dpopp07: I think you also want to check for if `self.authenticator` is null and throw an error if that's the case

ibm_cloud_sdk_core/cp4d_token_manager.py
```diff
@@ -0,0 +1,76 @@
...
+        :attr str username: The username
+        :attr str password: The password
+        :attr str url: The url for authentication
+        :attr bool disable_ssl_verification: enables/ disabled ssl verification
+        :attr dict headers: user-defined headers
+        :attr dict proxies: user-defined proxies
+        """
+        self.username = username
+        self.password = password
+        url = url + '/v1/preauth/validateAuth' if url else None
```
dpopp07: You should check if `url` already ends with the string `'/v1/preauth/validateAuth'` and only add that if it's not there.

ibm_cloud_sdk_core/iam_token_manager.py
```diff
@@ -39,40 +58,32 @@ def request_token(self):
...
         }
 
         # Use bx:bx as default auth header creds
         auth_tuple = ('bx', 'bx')
 
         # If both the clientId and secret were specified by the user, then use them
-        if self.iam_client_id and self.iam_client_secret:
-            auth_tuple = (self.iam_client_id, self.iam_client_secret)
+        if self.client_id and self.client_secret:
+            auth_tuple = (self.client_id, self.client_secret)
```
dpopp07: The new behavior is to just not send an auth header if the client id and secret are not set, rather than defaulting to `bx:bx`.

ibm_cloud_sdk_core/utils.py
```diff
@@ -50,3 +53,109 @@ def string_to_datetime(string):
...
+    Checks the credentials file and VCAP_SERVICES environment variable
+    :param service_name: The service name
+    :return: the authenticator
+    """
+    authenticator = None
+    # 1. Credentials from credential file
+    config = read_from_external_sources(service_name)
+    if config:
+        authenticator = contruct_authenticator(config)
+
```
dpopp07: After the credentials file, we should be looking at generic environment variables

ibm_cloud_sdk_core/utils.py
```diff
@@ -50,3 +53,109 @@ def string_to_datetime(string):
...
+                if vcap_service_credentials.get('username') and vcap_service_credentials.get('password'): # cf
+                    vcap_service_credentials['auth_type'] = 'basic'
+                elif vcap_service_credentials.get('apikey'):  # rc
+                    vcap_service_credentials['auth_type'] = 'iam'
+                else: # no other auth mechanism is supported
+                    vcap_service_credentials = None
+    return vcap_service_credentials
+
+def contruct_authenticator(config):
+    auth_type = config.get('auth_type')
```
dpopp07: If `auth_type` is not found, it should defaul to `iam`

README.md
ehdsouza: Thanks you :)

ibm_cloud_sdk_core/base_service.py
ehdsouza: The `self.authenticator` can be null initially when initializing the base class. In the generated class we would be doing the check to see if it is `None`

ibm_cloud_sdk_core/cp4d_token_manager.py
ehdsouza: Good suggestion!

ibm_cloud_sdk_core/base_service.py
```diff
@@ -14,292 +14,109 @@
...
+                 authenticator=None,
+                 disable_ssl_verification=False,
+                 display_name=None):
         """
-        It loads credentials with the following preference:
-        1) Credentials explicitly set in the request
-        2) Credentials loaded from credentials file if present
-        3) Credentials loaded from VCAP_SERVICES environment variable if available and use_vcap_services is True
+        :attr str url: The url for service api calls
+        :attr Athenticator authenticator: The authenticator for authentication
```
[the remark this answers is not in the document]
dpopp07: I think this still needs to be fixed!

ibm_cloud_sdk_core/iam_token_manager.py
```diff
@@ -81,5 +90,23 @@ def set_iam_authorization_info(self, iam_client_id, iam_client_secret):
         If these values are not supplied, then a default Authorization header
         is used.
```
dpopp07: Last little docs change, there is no default header anymore

ibm_cloud_sdk_core/utils.py
```diff
@@ -50,3 +53,126 @@ def string_to_datetime(string):
...
+    :param service_name: The service name
+    :return: the authenticator
+    """
+    authenticator = None
+    # 1. Credentials from credential file
+    config = read_from_credential_file(service_name)
+    if config:
+        authenticator = contruct_authenticator(config)
+
+    # 3. From env variables
```
dpopp07: Small typo
```suggestion
    # 2. From env variables
```

README.md
```diff
@@ -21,17 +21,52 @@ easy_install --upgrade ibm-cloud-sdk-core
...
+authenticator = BasicAuthenticator(<your_username>, <your_password>)
+```
+
+### IAM
+This indicates that IAM token authentication is to be used. Users can pass in a `apikey` and the SDK will generate a Bearer Auth header to send with requests to the service.
+
+```py
+from ibm_cloud_sdk_core.authenticators import IamAuthenticator
+
+authenticator = IAMAuthenticator(<your_apikey>)
```
dpopp07: Which one is it? `IamAuthenticator` or `IAMAuthenticator`? Both are used in this example

ibm_cloud_sdk_core/authenticators/no_auth_authenticator.py
```diff
@@ -0,0 +1,26 @@
...
+    authentication_type = 'noauth'
```
dpopp07: If we're changing everything to `NoAuth`, we should be consistent. This value should be `noAuth` as well.

ibm_cloud_sdk_core/utils.py
```diff
@@ -50,3 +53,126 @@ def string_to_datetime(string):
...
+            url=config.get('auth_url'),
+            disable_ssl_verification=config.get('auth_disable_ssl'))
+    elif auth_type == 'iam' and config.get('apikey'):
+        authenticator = IAMAuthenticator(
+            apikey=config.get('apikey'),
+            url=config.get('auth_url'),
+            client_id=config.get('client_id'),
+            client_secret=config.get('client_secret'),
+            disable_ssl_verification=config.get('auth_disable_ssl'))
+    elif auth_type == 'noauth':
```
dpopp07:
```suggestion
    elif auth_type == 'noAuth':
```

resources/ibm-credentials-no-auth.env
```diff
@@ -0,0 +1 @@
+WATSON_AUTH_TYPE=noauth
```
dpopp07:
```suggestion
WATSON_AUTH_TYPE=noAuth
```

PR #22, opened 2019-08-29
vcap and json loads changes

ehdsouza:
This PR supports the following:
* Convert data object to string before making a request
* Temporary handle vcap services... This is a temporary change as we pass in only `service name`. So a VCAP service like `compare-and-comply` would be converted to `compare_and_comply`

### DIFF
--- a/ibm_cloud_sdk_core/base_service.py
+++ b/ibm_cloud_sdk_core/base_service.py
@@ -44,7 +44,7 @@ class BaseService(object):
         """
         :attr str url: The url for service api calls
         :attr Authenticator authenticator: The authenticator for authentication
-        :attr bool disable_ssl_verification: enables/ disabled ssl verification
+        :attr bool disable_ssl_verification: enables/ disables ssl verification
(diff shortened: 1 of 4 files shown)


padamstx:
>     * Temporary handle vcap services... This is a temporary change as we pass in only `service name`. So a VCAP service like `compare-and-comply` would be converted to `compare_and_comply`

We shouldn't be doing any transformation on the service name when we use it to retrieve an entry in VCAP_SERVICES.   If the service name obtained from the API definition and inserted into the generated code is "compare-and-comply", then that's the key that we'd expect to find in VCAP_SERVICES.
We only do the transformation on the service name when looking for properties in the credential file or environment variables.   
There is a section on this in the Authentication design:
https://github.ibm.com/CloudEngineering/openapi-sdkgen/wiki/Core-SDK-Authentication-Design#a-note-about-service-names

ehdsouza:
@padamstx this was done temporary before the meeting as a candidate release had to get out

PR #23, opened 2019-08-30
Model conversion

ehdsouza:
Now since the `x-alternate-name` is not present in the api definitions of major release, we can remove the `_from_dict ` step to avoid potential errors received especially in import/ export workspace

### DIFF
--- a/ibm_cloud_sdk_core/base_service.py
+++ b/ibm_cloud_sdk_core/base_service.py
@@ -206,10 +206,6 @@ class BaseService(object):
 
     @staticmethod
     def _convert_model(val, classname=None):
-        if classname is not None and not hasattr(val, "_from_dict"):
-            if isinstance(val, str):
-                val = json_import.loads(val)
-            val = classname._from_dict(dict(val))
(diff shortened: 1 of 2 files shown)


ibm_cloud_sdk_core/base_service.py
```diff
@@ -206,10 +206,9 @@ def prepare_request(self, method, url, headers=None,
...
     @staticmethod
     def _convert_model(val, classname=None):
-        if classname is not None and not hasattr(val, "_from_dict"):
+        if classname is not None:
             if isinstance(val, str):
                 val = json_import.loads(val)
-            val = classname._from_dict(dict(val))
         if hasattr(val, "_to_dict"):
             return val._to_dict()
         return val
```
mkistler: I think this method could be even simpler, since there is no longer any need for the `classname` parameter:
```
    def _convert_model(val):
        if isinstance(val, str):
            val = json_import.loads(val)
        if hasattr(val, "_to_dict"):
            return val._to_dict()
        return val
```

PR #24, opened 2019-09-05
Logging and debugging

ehdsouza:
This PR enables logging and debugging

* The underlying http urllib3 library doesn't use the logging, but rather print statement to stdout. functions `enable_debugging()` and `disable_debugging` is used to support it
* We also need to provide users with python's logging capabilities:
   1) `enable_logging` takes in a config file path
   2) `get_logger ` allows users to configure loggers

readme is updated here: https://github.com/watson-developer-cloud/python-sdk/pull/687


ehdsouza:
This is no longer needed

PR #26, opened 2019-09-11
logging, service_url and auth_type case insensitive changes

ehdsouza:
This PR contains the following:
1) Show logging in readme
3) Change url to service_url
3)  Validate that service_url is set in the prepare_request
4) Make "auth type" case-insensitive


ehdsouza:
@padamstx there is already a test case covering it: https://github.com/IBM/python-sdk-core/blob/next/resources/ibm-credentials-no-auth.env

PR #28, opened 2019-09-16
Change order of credential path and read_external_sources and more

ehdsouza:
This PR does the following:
* Client properties like `disable_ssl` and `url` were read only from env, so like rest of the sdk's separated a `read_external_sources`
* Re-ordeing of fetching credentials from working directory before home dir
* Also removed setting of client properties in the constructor as they now go in the generated service classes

related to https://github.ibm.com/Watson/developer-experience/issues/7521


ibm_cloud_sdk_core/utils.py
```diff
@@ -54,29 +54,40 @@ def string_to_datetime(string):
...
+    1. Credentials file(ibm-credentials.env)
+    2. Environment variables
+    3. VCAP Services(Cloud Foundry)
+    :param service_name: The service name
+    :return: the authenticator
+    """
+    authenticator = None
+    config = read_external_sources(service_name)
+    if config:
+        authenticator = contruct_authenticator(config)
```
padamstx: s/contruct/construct/

ibm_cloud_sdk_core/base_service.py
```diff
@@ -65,7 +65,7 @@ def __init__(self,
 
         if display_name:
             service_name = display_name.replace(' ', '_').lower()
```
padamstx: I think we also need to replace '-' with '_' in the service_name value.    Some values for the vcap services name have -'s in them.

ibm_cloud_sdk_core/utils.py
padamstx: Also, "construct_authenticator" could probably be private since it should only be called from inside the get_authenticator_from_environment public method.

ibm_cloud_sdk_core/utils.py
```diff
@@ -100,16 +111,16 @@ def read_from_credential_file(service_name, separator='='):
     # File path specified by an env variable
     credential_file_path = getenv('IBM_CREDENTIALS_FILE')
 
-    # Home directory
+    # Top-level of the project directory
```
padamstx: This is actually the "Current working directory", right?

ibm_cloud_sdk_core/utils.py
```diff
@@ -100,16 +111,16 @@ def read_from_credential_file(service_name, separator='='):
...
     if credential_file_path is None:
-        file_path = join(expanduser('~'), DEFAULT_CREDENTIALS_FILE_NAME)
+        file_path = join(
+            dirname(dirname(abspath(__file__))), DEFAULT_CREDENTIALS_FILE_NAME)
         if isfile(file_path):
             credential_file_path = file_path
 
-    # Top-level of the project directory
+    # Home directory
```
padamstx: In the other cores, we look in the "unix" home directory, as well as two different "home" directories associated with the Windows OS.
You can see that here in the Java SDK:
https://github.com/IBM/java-sdk-core/blob/e3d5330b0d9e2ef9cf084cb7c6aa97d6ae4f1fb4/src/main/java/com/ibm/cloud/sdk/core/util/CredentialUtils.java#L208

Does `expanduser('~')` handle all three of those?   If not, then we probably need to add the missing steps.

ibm_cloud_sdk_core/utils.py
```diff
@@ -125,15 +136,15 @@ def read_from_credential_file(service_name, separator='='):
     return config
 
 def _parse_key_and_update_config(config, service_name, key, value):
-    if service_name in key:
+    if key.startswith(service_name):
         index = key.find('_')
```
padamstx: I don't think this works if the service_name contains an "_" does it???

You can see the logic that's used by the java core in `CredentialUtils.parseCredentials()`, to parse a line from either a credential file or an env var setting, which looks like this:
```
serviceName = serviceName.toUpperCase().replaceAll("-", "_") + "_";
...
if (key.startsWith(serviceName)) {
  String credentialName = key.substring(serviceName.length());
  ...
}
```

ehdsouza: This will work as before calling the `_parse_key_and_update_config `, the calling function perform's the replace

ibm_cloud_sdk_core/utils.py
ehdsouza: yes https://docs.python.org/3/library/os.path.html#os.path.expanduser

ibm_cloud_sdk_core/utils.py
padamstx: The service_name passed into `_parse_key_and_update_config` could have an "_" in it, so this code fragment won't work:
```
    if key.startswith(service_name):
        index = key.find('_')
        if index != -1:
            config[key[index + 1:]] = value
```
If this code is processing a key of "personality_insights_url", you'd end up with a config map with this entry:
```
   "insights_url" --> <url value>
```
Right?

Please add a test that includes a multi-word service name and you will find out exactly what happens.

resources/ibm-credentials-iam.env
```diff
@@ -1,2 +1,4 @@
 WATSON_APIKEY=5678efgh
-WATSON_AUTH_TYPE=iam
\ No newline at end of file
+WATSON_AUTH_TYPE=iam
+WATSON_URL=https://gateway-s.watsonplatform.net/watson/api
+WATSON_DISABLE_SSL=False
```
padamstx: Add some test properties that include a multi-word service name, rather than just "WATSON".

ibm_cloud_sdk_core/utils.py
ehdsouza: Ahaaa I see that the truncation happening here! Good catch! 👍

ehdsouza:
@padamstx addressed the multi word service name issue.

resources/ibm-credentials-iam.env
mediumTaj: Erika can you add these tests properties?

ibm_cloud_sdk_core/utils.py
```diff
@@ -125,15 +134,17 @@ def read_from_credential_file(service_name, separator='='):
     return config
 
 def _parse_key_and_update_config(config, service_name, key, value):
-    if service_name in key:
-        index = key.find('_')
+    service_name = service_name.replace(' ', '_').replace('-', '_').lower()
+    key = key.replace(' ', '_').replace('-', '_').lower()
+    if key.startswith(service_name):
+        index = key.find(service_name)
```
padamstx: since we entered this block because `key` starts with `service_name`, you shouldn't need to do a find at all, right?
since the key starts with the service_name the index should always be 0 (or 1, whatever index the first character has).
Please see the logic in the CredentialUtils.parseCredentials() function:
https://github.com/IBM/java-sdk-core/blob/e3d5330b0d9e2ef9cf084cb7c6aa97d6ae4f1fb4/src/main/java/com/ibm/cloud/sdk/core/util/CredentialUtils.java#L352
Specifically this:
```
   serviceName = serviceName.toUpperCase().replaceAll("-", "_") + "_";
   ...
      if (key.startsWith(serviceName)) {
        String credentialName = key.substring(serviceName.length());
```
Note that this function changes the serviceName value to do the character replacement AND adds a `"_"` on the end so that the updated value of serviceName (e.g. `PERSONALITY_INSIGHTS_`) will serve as the actual "prefix" that we'd expect to see in property and env variable names.   Because we add the trailing `"_"` on the end, there's no `+1` needed when we compute the length of serviceName.

ibm_cloud_sdk_core/base_service.py
```diff
@@ -63,15 +63,6 @@ def __init__(self,
             raise ValueError(
                 'authenticator should be of type Authenticator')
 
-        if display_name:
-            service_name = display_name.replace(' ', '_').lower()
-            config = read_from_env_variables(service_name)
-            if config.get('url'):
```
ehdsouza: setting of client properties go in the generated service classes

ibm_cloud_sdk_core/utils.py
```diff
@@ -125,15 +134,17 @@ def read_from_credential_file(service_name, separator='='):
...
         if index != -1:
-            config[key[index + 1:]] = value
+            config[key[index + len(service_name) + 1:]] = value
 
 def read_from_vcap_services(service_name):
-    service_name = service_name.replace(' ', '_').lower()
+    service_name = service_name.replace(' ', '_').replace('-', '_').lower()
```
padamstx: When we get config info from VCAP_SERVICES, we should not do any transformations on `service_name` at all.
This is because the exact service name is configured in the API definition and is inserted into the generated code and should be used as-is.   So, a vcap services name obtained from the API definition of `personality-insights` is the exact string that we'd expect to find within VCAP_SERVICES.

ibm_cloud_sdk_core/utils.py
```diff
@@ -125,15 +134,17 @@ def read_from_credential_file(service_name, separator='='):
     return config
 
 def _parse_key_and_update_config(config, service_name, key, value):
-    if service_name in key:
-        index = key.find('_')
+    service_name = service_name.replace(' ', '_').replace('-', '_').lower()
+    key = key.replace(' ', '_').replace('-', '_').lower()
```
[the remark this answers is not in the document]
ehdsouza: But for consistency isn't it better to have the key and service_name both converted?

What if a user has a environment variable like `IBM-WATSON_USERNAME`?

ibm_cloud_sdk_core/utils.py
ehdsouza: Agree that I don't need to use find to get index, but how can we guarantee that the key would always be uppercase? I would prefer we transform both so that we perform the correct string matching

ibm_cloud_sdk_core/utils.py
ehdsouza: I perform the key conversion here: https://github.com/IBM/python-sdk-core/pull/28/files#diff-7048c45d573bb78c4bac9b0291b4e065R152 so that when service_name = `personality-insights` and vcap name is `personality-insights` would both be converted to `personality_insights`

ibm_cloud_sdk_core/utils.py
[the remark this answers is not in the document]
padamstx: If the user doesn't follow the conventions, there's not much we can do about that.   My opinion is that we shouldn't be changing the configuration data (cred file properties/env variables/VCAP_SERVICES) that the user has defined.

ibm_cloud_sdk_core/utils.py
padamstx: Why?   All you need to do is compare the "as-is" service_name as it was generated in the service and compare it with the outer-most map key within the VCAP_SERVICES json object.   By trying to do all these extra, unnecessary transformations, you're actually accepting a different "language" than the other cores.

ibm_cloud_sdk_core/utils.py
```diff
@@ -119,62 +128,62 @@ def read_from_credential_file(service_name, separator='='):
...
+                        vcap_service_credentials['APIKEY'] = vcap_service_credentials.get('apikey')
                     else: # no other auth mechanism is supported
-                        vcap_service_credentials = None
+                        vcap_service_credentials = {}
     return vcap_service_credentials
 
-def contruct_authenticator(config):
-    auth_type = config.get('auth_type').lower() if config.get('auth_type') else 'iam'
+def _construct_authenticator(config):
+    auth_type = config.get('AUTH_TYPE').lower() if config.get('AUTH_TYPE') else 'iam'
```
padamstx: I didn't comment on this before, and I don't think we need to change anything right now, but... when we implement the service config feature, it might be good to define constants for the various config properties rather than use the literal strings like this.

PR #30, opened 2019-09-27
fix: support duplicate part names in 'files' parameter of prepare_req…

padamstx:
Fixes: https://github.ibm.com/arf/planning-sdk-squad/issues/1064

This PR allows the "prepare_request" caller to pass in a list of tuples (e.g. `[(part-name, (file-tuple)), ...]`) for the 'files' parameter instead of a dictionary.   Each list item will result in a separate part to be created in the multi-part form.   This change allows multiple list items to use the same part-name.

### DIFF
--- a/ibm_cloud_sdk_core/base_service.py
+++ b/ibm_cloud_sdk_core/base_service.py
@@ -197,16 +197,26 @@ class BaseService(object):
         if self.authenticator:
             self.authenticator.authenticate(request)
 
-        files = remove_null_values(files)
+        # Next, we need to process the 'files' argument to try to fill in
+        # any missing filenames where possible.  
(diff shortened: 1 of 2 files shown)


ibm_cloud_sdk_core/base_service.py
```diff
@@ -197,16 +197,26 @@ def prepare_request(self, method, url, headers=None,
...
+        new_files = []
         if files is not None:
-            for k, file_tuple in files.items():
-                if file_tuple and len(
-                        file_tuple) == 3 and file_tuple[0] is None:
-                    file = file_tuple[1]
+            # If 'files' is a dictionary, transform it into a list of tuples.
+            if isinstance(files, dict):
+                files = remove_null_values(files)
+                files = files.items()
```
ehdsouza: Nice 👍

PR #35, opened 2019-10-18
Document python core

mgrattray:
I followed the [PEP 257](https://www.python.org/dev/peps/pep-0257/) Google Docstring [style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).

Other documentation: https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings

Issue: https://github.ibm.com/arf/planning-sdk-squad/issues/1117

### DIFF
--- a/ibm_cloud_sdk_core/__init__.py
+++ b/ibm_cloud_sdk_core/__init__.py
@@ -12,6 +12,22 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+"""Classes and helper functions used by generated SDKs.
+
+classes:
+    BaseService: Abstract class for common functionality between each service.
+    DetailedResponse: The object returned from successful service operations.
+    IAMTokenManager: Requests and refreshes IAM tokens using an apikey, and optionally a client_id and client_secret pair.
+    JWTTokenManager: Abstract class for common functionality between each JWT token manager.
+    CP4DTokenManager: Requests and refreshes CP4D tokens given a username and password.
+    ApiException: Custom exception class for errors returned from service operations.
+
+functions:
+    datetime_to_string: Serializes a datetime to a string.
+    string_to_datetime: De-serializes a string to a datetime.
+    read_external_sources: Get config object from external sources.
+    get_authenticator_from_environment: Get authenticator from external sources.
+"""
 
 from .base_service import BaseService
 from .detailed_response import DetailedResponse
@@ -19,4 +35,5 @@ from .iam_token_manager import IAMTokenManager
 from .jwt_token_manager import JWTTokenManager
 from .cp4d_token_manager import CP4DTokenManager
 from .api_exception import ApiException
-from .utils import datetime_to_string, string_to_datetime, get_authenticator_from_environment, read_external_sources
+from .utils import datetime_to_string, string_to_datetime, read_external_sources
+from .get_authenticator import get_authenticator_from_environment
--- a/ibm_cloud_sdk_core/api_exception.py
+++ b/ibm_cloud_sdk_core/api_exception.py
@@ -13,17 +13,26 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from requests import Response
+from typing import Optional
+
 
 class ApiException(Exception):
-    """
-    Custom exception class for errors returned from APIs.
+    """Custom exception class for errors returned from operations.
 
-    :param int code: The HTTP status code returned.
-    :param str message: A message describing the error.
-    :param dict info: A dictionary of additional information about the error.
-    :param response http_response: response
+    Args:
+        code: HTTP status code of the error response.
+        message: The error response body. Defaults to None.
+        http_response: The HTTP response of the failed request. Defaults to None.
+
+    Attributes:
+        code (int): HTTP status code of the error response.
+        message (str): The error response body.
+        http_response (requests.Response): The HTTP response of the failed request.
+        global_transaction_id (str, optional): Globally unique id the service endpoint has given a transaction.
     """
-    def __init__(self, code, message=None, info=None, http_response=None):
+
+    def __init__(self, code: int, message: Optional[str] = None, http_response: Optional[Response] = None):
         # Call the base class constructor with the parameters it needs
         super(ApiException, self).__init__(message)
         self.message = message
@@ -40,12 +49,7 @@ class ApiException(Exception):
             msg += ' , X-global-transaction-id: ' + str(self.global_transaction_id)
         return  msg
 
-    def _get_error_message(self, response):
-        """
-        Gets the error message from a JSON response.
-        :return: the error message
-        :rtype: string
-        """
+    def _get_error_message(self, response: Response):
         error_message = 'Unknown error'
         try:
             error_json = response.json()
--- a/ibm_cloud_sdk_core/authenticators/__init__.py
+++ b/ibm_cloud_sdk_core/authenticators/__init__.py
@@ -13,6 +13,25 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+"""The ibm_cloud_sdk_core project supports the following types of authentication:
+
+  Basic Authentication
+  Bearer Token
+  Identity and Access Management (IAM)
+  Cloud Pak for Data
+  No Authentication
+
+  The authentication types that are appropriate for a particular service may vary from service to service.
+  Each authentication type is implemented as an Authenticator for consumption by a service.
+
+classes:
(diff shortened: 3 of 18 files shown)


ibm_cloud_sdk_core/__init__.py
```diff
@@ -12,11 +12,27 @@
...
+"""
 
 from .base_service import BaseService
 from .detailed_response import DetailedResponse
 from .iam_token_manager import IAMTokenManager
 from .jwt_token_manager import JWTTokenManager
 from .cp4d_token_manager import CP4DTokenManager
 from .api_exception import ApiException
-from .utils import datetime_to_string, string_to_datetime, get_authenticator_from_environment, read_external_sources
+from .utils import datetime_to_string, string_to_datetime, read_external_sources
```
ehdsouza: Is there a reason to remove `get_authenticator_from_environment`? It is used in the watson SDKs: https://github.com/watson-developer-cloud/python-sdk/blob/master/ibm_watson/assistant_v1.py#L29

ibm_cloud_sdk_core/api_exception.py
```diff
@@ -13,17 +13,26 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import requests
```
ehdsouza: you can import just the `Response`, just like below for `Optional`

ibm_cloud_sdk_core/api_exception.py
```diff
@@ -13,17 +13,26 @@
...
-    """
-    Custom exception class for errors returned from APIs.
+    """Custom exception class for errors returned from operations.
 
-    :param int code: The HTTP status code returned.
-    :param str message: A message describing the error.
-    :param dict info: A dictionary of additional information about the error.
-    :param response http_response: response
+    Args:
+        code: HTTP status code of the error response.
```
ehdsouza: No specifying type here?

ibm_cloud_sdk_core/authenticators/authenticator.py
```diff
@@ -17,16 +17,20 @@
 from abc import ABC, abstractmethod
 
 class Authenticator(ABC):
+    """This interface defines the common methods and constants associated with an Authenticator implementation.
```
ehdsouza: I think you can close this string in the same line

ibm_cloud_sdk_core/authenticators/basic_authenticator.py
```diff
@@ -17,23 +17,37 @@
 from .authenticator import Authenticator
 from ..utils import has_bad_first_or_last_char
 import base64
+import requests
```
ehdsouza: How about importing what you need directly?

ibm_cloud_sdk_core/authenticators/no_auth_authenticator.py
```diff
@@ -17,6 +17,8 @@
 from .authenticator import Authenticator
 
 class NoAuthAuthenticator(Authenticator):
+    """Performs no authentication.
```
ehdsouza: Can close in same line

ibm_cloud_sdk_core/__init__.py
mgrattray: @ehdsouza I added it back to https://github.com/IBM/python-sdk-core/pull/35/files#diff-273d7ea91f33cc339a1dcb6d2568b5adR44. I moved it because there were circular dependencies that were interfering with proper type hinting.

ibm_cloud_sdk_core/api_exception.py
mgrattray: 👍

ibm_cloud_sdk_core/api_exception.py
mgrattray: AFAIK you do not have to specify type for args as long as it is hinted.

ibm_cloud_sdk_core/authenticators/basic_authenticator.py
mgrattray: 👍

ibm_cloud_sdk_core/authenticators/authenticator.py
mgrattray: 👍

ibm_cloud_sdk_core/authenticators/no_auth_authenticator.py
mgrattray: 👍

mkistler:
It is surprising to me the adding documentation would be a breaking change.  Can you explain that?

ibm_cloud_sdk_core/__init__.py
mkistler: But there is now a dissonance here between the comment listing the classes and functions exported (I love the comment!  Thanks for adding that) and the actual exports in this file.  So perhaps you could add a comment here to explain with that one method isn't exported here, and where it is exported from instead.

ibm_cloud_sdk_core/api_exception.py
```diff
@@ -13,17 +13,26 @@
...
+        message: The error response body. Defaults to None.
+        http_response: The HTTP response of the failed request. Defaults to None.
+
+    Attributes:
+        code (int): HTTP status code of the error response.
+        message (str): The error response body.
+        http_response (requests.Response): The HTTP response of the failed request.
+        global_transaction_id (str, optional): Globally unique id the service endpoint has given a transaction.
```
mkistler: I was a little surprised by the change in documentation style, but then discovered that this new style is exactly what our "CONTRIBUTING" doc says to use.  So nice job in making us adhere to our own style guide!

mgrattray:
I thought it would a breaking change since I moved the `get_authenticator_from_environment` method.

So now code that did:
`from ibm_cloud_sdk_core import get_authenticator_from_environment`

Will be broken and be have to be updated too:
`from ibm_cloud_sdk_core.authenticators import get_authenticator_from_environment`

However, that's not a breaking change for the generator. Only users who previously relied upon the package.

ibm_cloud_sdk_core/__init__.py
mgrattray: I've fixed the comment.

ibm_cloud_sdk_core/api_exception.py
mgrattray: 👍

ibm_cloud_sdk_core/authenticators/__init__.py
```diff
@@ -13,10 +13,82 @@
...
 
 from .authenticator import Authenticator
 from .basic_authenticator import BasicAuthenticator
 from .bearer_token_authenticator import BearerTokenAuthenticator
 from .cp4d_authenticator import CloudPakForDataAuthenticator
 from .iam_authenticator import IAMAuthenticator
 from .no_auth_authenticator import NoAuthAuthenticator
+from ..utils import read_external_sources
+
+def get_authenticator_from_environment(service_name: str) -> Authenticator:
```
padamstx: As we discussed, I think we should try to expose the `get_authenticator_from_environment` function in a way that doesn't create an incompatibility with existing users of the python core v1.0.0.

ibm_cloud_sdk_core/authenticators/basic_authenticator.py
```diff
@@ -44,9 +58,16 @@ def validate(self):
...
+
+        Basic Authorization will be added to the request's headers in the form:
+
+            Authorization: Basic <encoded username and password>
+
+        Args:
+            req: The request to add basic auth information too. Must contain a key to a dictionary
+            called headers.
         """
         authstring = "{0}:{1}".format(self.username, self.password)
```
padamstx: This might be a little beyond the scope of your task, but there's really no reason why we can't cache the Authorization header value (i.e. update it any time the username/password is set or updated), then just use the cached value inside the `authenticate()` method.

ibm_cloud_sdk_core/base_service.py
```diff
@@ -33,23 +34,39 @@
 # import http.client as http_client
 # http_client.HTTPConnection.debuglevel = 1
 
-
 class BaseService(object):
-
+    """Abstract class for common functionality between generated services.
```
padamstx: Is this really an "abstract" class???

padamstx: Also, maybe a comment like "Common functionality shared by generated service classes." would be a little bit more clear?

ibm_cloud_sdk_core/authenticators/__init__.py
mgrattray: this is not the latest and has been fixed.

ibm_cloud_sdk_core/authenticators/basic_authenticator.py
mgrattray: 👍

ibm_cloud_sdk_core/base_service.py
mgrattray: :+1:

mgrattray:
@padamstx It's been tested with and is ready for review.

PR #40, opened 2019-10-27
Fix linting for Python3 and fix all lint issues

mkistler:
This PR fixes the linting script so that it will run with Python3 -- now the only version supported by the core.  To make the set of changes manageable, I disabled a few of the less serious lint rules.  Most of the changes were to conform to these linting rules:

- Line too long
- standard import should be placed before third-party import
- third-party import should be placed before package import
- Class 'BaseService' inherits from object, can be safely removed from bases in python3
- Redefining built-in 'tuple'
- Method could be a function
- Consider using a dictionary comprehension

I also took this opportunity to add external methods for `convert_model`, `convert_list`, and `encode_path_vars`. so that users of the core can use these methods rather than the corresponding methods that start with "_".


ibm_cloud_sdk_core/base_service.py
```diff
@@ -308,16 +310,32 @@ def prepare_request(self,
...
                         filename = basename(file.name)
-                        tuple = (filename, tuple[1], tuple[2])
-                new_files.append((part_name, tuple))
+                        file_tuple = (filename, file_tuple[1], file_tuple[2])
+                new_files.append((part_name, file_tuple))
         request['files'] = new_files
         return request
 
+    @staticmethod
+    def encode_path_vars(*args):
```
mkistler: This method should probably have type-hinting.  Should it be?
```
def encode_path_vars(*args: List[str]) -> List[str] :
```

ibm_cloud_sdk_core/utils.py
```diff
@@ -88,6 +89,32 @@ def string_to_datetime(string: str) -> datetime.datetime:
     """
     return date_parser.parse(string)
 
+def convert_model(val):
```
mkistler: I'd like to add type-hints on this method but not sure how to do that.  The input `val` could be anything.  The output will be a `dict` if the input had a `to_dict` method, but otherwise could be anything.  So is the right answer:
```
def convert_model(val: Any) -> Any:
```
and if yes, then is it even worth providing type hints?

ibm_cloud_sdk_core/utils.py
```diff
@@ -88,6 +89,32 @@ def string_to_datetime(string: str) -> datetime.datetime:
...
+        val: An object
+
+    Returns:
+        A dict representation of the input object.
+    """
+    if hasattr(val, "to_dict"):
+        return val.to_dict()
+    return val
+
+def convert_list(val):
```
mkistler: Similar dilemma with type hints on this method.

mkistler:
I will fix the coverage so that the PR checks pass.  We should probably add pylint to PR checks as well.

ibm_cloud_sdk_core/utils.py
```diff
@@ -88,6 +89,32 @@ def string_to_datetime(string: str) -> datetime.datetime:
...
     return date_parser.parse(string)
 
+def convert_model(val):
+    """Convert a model object into an equivalent dict.
+
+    Arguments:
+        val: An object
+
+    Returns:
+        A dict representation of the input object.
```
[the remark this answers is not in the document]
mgrattray: I do not think we need to use `Union` here anymore. it can just be: `def convert_model(val:  any) -> dict:`

mkistler:
@mgrattray I decided to back off on raising the ValueError until the next major release, just for safety.

PR #41, opened 2019-11-04
Update vcap external config parsing

rmkeezer:
Change vcap service external config parsing to work like new behavior here: https://github.ibm.com/CloudEngineering/openapi-sdkgen/wiki/Design-Of-Service-Config-Changes#external-config-behavior

See https://github.ibm.com/arf/planning-sdk-squad/issues/1100


ibm_cloud_sdk_core/utils.py
```diff
@@ -22,6 +23,9 @@
 
 import dateutil.parser as date_parser
 
+if 'VCAP_SERVICES_FILE' not in os.environ.keys():
```
padamstx: I don't think we want to support an env variable that specifies the name of a file containing the VCAP object.
The only thing we currently support is the VCAP_SERVICES environment variable that contains the actual VCAP object (JSON string).
On the surface, adding support for loading the VCAP object from a file might be a good thing, but:
1) We really don't want to encourage anyone to use VCAP_SERVICES if they aren't already using it (it's sort of "stale").
2) We'd need to replicate this new functionality across all the cores.

ibm_cloud_sdk_core/utils.py
```diff
@@ -218,21 +222,30 @@ def __read_from_vcap_services(service_name: str) -> dict:
         A set of service configuration key-value pairs.
     """
     vcap_services = getenv('VCAP_SERVICES')
+    if not vcap_services:
+        vcap_services_file = getenv('VCAP_SERVICES_FILE')
```
padamstx: let's remove the code that loads the VCAP object from a file.

ibm_cloud_sdk_core/utils.py
```diff
@@ -218,21 +222,30 @@ def __read_from_vcap_services(service_name: str) -> dict:
...
-                        vcap_service_credentials['AUTH_TYPE'] = 'basic'
-                        vcap_service_credentials['USERNAME'] = vcap_service_credentials.get('username')
-                        vcap_service_credentials['PASSWORD'] = vcap_service_credentials.get('password')
-                    elif vcap_service_credentials.get('apikey'):  # rc
-                        vcap_service_credentials['AUTH_TYPE'] = 'iam'
-                        vcap_service_credentials['APIKEY'] = vcap_service_credentials.get('apikey')
-                    else: # no other auth mechanism is supported
-                        vcap_service_credentials = {}
+            for i in range(len(services[key])):
+                if vcap_service_credentials and isinstance(vcap_service_credentials, dict):
```
padamstx: at this point, we haven't yet set `vcap_service_credentials` to anything, plus it's initialized above so this `if` statement would always be true, wouldn't it?

ibm_cloud_sdk_core/utils.py
```diff
@@ -218,21 +222,30 @@ def __read_from_vcap_services(service_name: str) -> dict:
...
+                    break
+                if services[key][i].get('name') == service_name:
+                    vcap_service_credentials = services[key][i].get('credentials')
+        if not vcap_service_credentials:
+            if service_name in services.keys():
+                vcap_service_credentials = services.get(service_name)[0].get('credentials')
```
padamstx: this seems like a lot of clever coding shortcuts, which is fine, but what if the VCAP_SERVICES object contains the desired service key, but yet it maps to an empty service list?
Let's make sure we have tests that simulate a bunch of odd scenarios, like the service key maps to an empty list, or it maps to an entry that doesn't contain a "credentials" field, etc.

ibm_cloud_sdk_core/utils.py
```diff
@@ -218,21 +222,30 @@ def __read_from_vcap_services(service_name: str) -> dict:
...
+
+        if vcap_service_credentials and isinstance(vcap_service_credentials, dict):
+            if vcap_service_credentials.get('username') and vcap_service_credentials.get('password'): # cf
```
padamstx: I'm really confused by this... we seem to be returning `vcap_service_credentials` such that it will contain the "credentials" field of the service entry (as a dictionary... e.g. "username", "password", "apikey", "url", etc.) plus all the additional properties that we explicitly set in there (e.g. "USERNAME", "PASSWORD, "IAM_APIKEY", "URL", etc.).
I think we should return a dictionary that contains only the well-defined properties (uppercase names) that will be used to either configure a service or construct an authenticator.

test/test_utils.py
```diff
@@ -136,6 +137,93 @@ def test_vcap_credentials():
     assert authenticator is None
     del os.environ['VCAP_SERVICES']
 
+    vcap_services = '{"test":[{"name": "testname",\
+        "credentials":{ \
```
padamstx: In order to simplify the tests and make it easy to add new vcap-related tests in the future, I think it would be good to capture all of the VCAP_SERVICES testdata perhaps in a file, then just read in that file and set the VCAP_SERVICES env var to contain the contents of the file, similar to the Go core tests.
In fact, you could probably just copy the same file and check it into the python core.

test/test_utils.py
```diff
@@ -136,6 +137,93 @@ def test_vcap_credentials():
...
+        "url":"https://gateway.watsonplatform.net/compare-comply/api",\
+        "username":"bogus username", \
+        "password":"bogus password"}}]}'
+
+    os.environ['VCAP_SERVICES'] = vcap_services
+    authenticator = get_authenticator_from_environment('testname')
+    assert authenticator is not None
```
padamstx: In these situations, we should also verify that we got back the correct Authenticator type:
```
assert isinstance(authenticator, BasicAuthenticator)
```

ibm_cloud_sdk_core/utils.py
rmkeezer: {} is falsy so it would only be true when a vcap_service_credentials is true, once found I need to break out of both loops.

ibm_cloud_sdk_core/utils.py
rmkeezer: I copied this logic from above so I didn't look too much into but I agree with your suggestion. I made it so only the well-defined properties are returned

test/test_utils.py
rmkeezer: done

test/test_utils.py
rmkeezer: I think it might be simplier to leave them as json snippets in the code like this since each one is slight different for each different test. If we put them in a file we'd need to lable each one to each test they go to and parse them from the file correct?

ibm_cloud_sdk_core/utils.py
rmkeezer: I added some more checks and it should return "{}" when finding an empty service or not finding credentials.

ibm_cloud_sdk_core/utils.py
rmkeezer: done

ibm_cloud_sdk_core/utils.py
```diff
@@ -235,17 +227,22 @@ def __read_from_vcap_services(service_name: str) -> dict:
...
-                vcap_service_credentials['PASSWORD'] = vcap_service_credentials.get('password')
+                new_vcap_creds['AUTH_TYPE'] = 'basic'
+                new_vcap_creds['USERNAME'] = vcap_service_credentials.get('username')
+                new_vcap_creds['PASSWORD'] = vcap_service_credentials.get('password')
+                vcap_service_credentials = new_vcap_creds
             elif vcap_service_credentials.get('apikey'):  # rc
-                vcap_service_credentials['AUTH_TYPE'] = 'iam'
-                vcap_service_credentials['APIKEY'] = vcap_service_credentials.get('apikey')
+                new_vcap_creds['AUTH_TYPE'] = 'iam'
+                new_vcap_creds['APIKEY'] = vcap_service_credentials.get('apikey')
```
padamstx: One minor detail... if the VCAP_SERVICES entry that we're copying values from ALSO contains an "iam_apikey" field, then we want to use that instead of the "apikey" field.   So "iam_apikey" trumps "apikey".   Whichever one we find, it should be copied into the new_vcap_creds['APIKEY'] entry in the dictionary.
If you look at the "v0.5.2" version of base_service.py in the python core repo you can see the logic that used to be there in terms of copying values from the vcap services entry.

PR #57, opened 2020-02-23
Pace requests to token server for new auth tokens

mkistler:
Implement pacing of token requests for both the initial/expired token case and for refresh requests.  These are implemented slightly differently because for token refresh we can allow requests to continue with the current token while it is not expired, but when the token expires all requests need to block behind the token request.


ibm_cloud_sdk_core/jwt_token_manager.py
```diff
@@ -80,6 +92,42 @@ def set_disable_ssl_verification(self, status: bool = False):
...
+                if not request_active:
+                    self.request_time = current_time
+
+            if not request_active:
+                token_response = self.request_token()
+                self._save_token_info(token_response)
+                self.request_time = 0
+                return
+
+            time.sleep(1)  # Sleep for 1 second before checking token again
```
ricellis: I think this might be a little on the long side, I can do a token exchange in 220 ms (with 18 ms ping time from my local laptop and I've seen ping times < 1 ms when running in Cloud). I think 0.1 might be closer to the mark if we don't want to bottleneck for any longer than necessary.

mkistler: This is actually a big improvement from the previous version of the code, where we might have slept for up to 60 seconds before checking again for a valid token 😄 .

I'm not really that concerned about a small bottleneck here, since it should be a relatively rare occurrence that we ever execute this code.  But I'm willing to cut it half, to 0.5s, to make it just a little more snappy.

PR #62, opened 2020-03-30
[Breaking] Only override content-type if it is none

rmkeezer:
ref: https://github.ibm.com/arf/planning-sdk-squad/issues/1201

Currently the header `content-type` is unconditionally overridden to "application/json". This change will only add it if the header is not set.

### DIFF
--- a/ibm_cloud_sdk_core/base_service.py
+++ b/ibm_cloud_sdk_core/base_service.py
@@ -290,7 +290,8 @@ class BaseService:
 
         if data and isinstance(data, dict):
             data = remove_null_values(data)
-            headers.update({'content-type': 'application/json'})
+            if headers.get('content-type') is None:
+                headers.update({'content-type': 'application/json'})
             data = json_import.dumps(data)
         request['data'] = data
 


ibm_cloud_sdk_core/base_service.py
```diff
@@ -290,7 +290,8 @@ def prepare_request(self,
 
         if data and isinstance(data, dict):
             data = remove_null_values(data)
-            headers.update({'content-type': 'application/json'})
+            if headers.get('content-type') is None:
```
mkistler: There is a subtle problem here -- header names are actually _case insensitive_.  So you can't just check for `content-type` -- the user might have passed `Content-type`.

rmkeezer: Wow I didn't know about that. I'll make this case insensitive. Thanks!

rmkeezer: It looks like headers is a CaseInsensitiveDict instead of a regular dict, so it already makes all keys lowercase including the .get() function. So if the user passes in `Content-type` it still works here.

padamstx: Just curious... what if the content-type is something other than "application/json" (i.e. "application/octet-stream") but yet "data" is a dict?   It looks like we'd send the JSON stringified version of the dict in the request body, along with content-type="application/octet-stream".   Not 100% if we could get ourselves into this situation, but if so, we should probably decide what to do in that case.   For example, maybe we should have logic such that if "data" is a dict and the content-type is not None but also not a JSON-like value (includes json-patch, etc.) then we throw an error or a warning or something?

PR #65, opened 2020-03-31
Add type annotations to parameters and return values

rmkeezer:
ref: https://github.ibm.com/arf/planning-sdk-squad/issues/1598

Add Python 3.5 type annotations to parameters and return values.

I merged PR #61 into this branch so it includes those changes until I rebase when it is merged into `v2`.

### DIFF
--- a/ibm_cloud_sdk_core/authenticators/authenticator.py
+++ b/ibm_cloud_sdk_core/authenticators/authenticator.py
@@ -22,6 +22,9 @@ class Authenticator(ABC):
     def authenticate(self, req: dict) -> None:
         """Perform the necessary authentication steps for the specified request.
 
+        Attributes:
+            req (dict): Will be modified to contain the appropriate authentication information.
+
         To be implemented by subclasses.
         """
         pass
@@ -29,6 +32,9 @@ class Authenticator(ABC):
     @abstractmethod
     def validate(self) -> None:
         """Validates the current set of configuration information in the Authenticator.
+
+        Raises:
+            ValueError: The configuration information is not valid for service operations.
 
         To be implemented by subclasses.
         """
--- a/ibm_cloud_sdk_core/authenticators/authenticator.py
+++ b/ibm_cloud_sdk_core/authenticators/authenticator.py
@@ -19,7 +19,7 @@ from abc import ABC, abstractmethod
 class Authenticator(ABC):
     """This interface defines the common methods and constants associated with an Authenticator implementation."""
     @abstractmethod
-    def authenticate(self, req: dict):
+    def authenticate(self, req: dict) -> None:
         """Perform the necessary authentication steps for the specified request.
 
         To be implemented by subclasses.
@@ -27,7 +27,7 @@ class Authenticator(ABC):
         pass
 
     @abstractmethod
-    def validate(self):
+    def validate(self) -> None:
         """Validates the current set of configuration information in the Authenticator.
 
         To be implemented by subclasses.
--- a/ibm_cloud_sdk_core/authenticators/basic_authenticator.py
+++ b/ibm_cloud_sdk_core/authenticators/basic_authenticator.py
@@ -62,7 +62,7 @@ class BasicAuthenticator(Authenticator):
                 'Please remove any surrounding {, }, or \" characters.')
 
 
-    def __construct_basic_auth_header(self) -> None:
+    def __construct_basic_auth_header(self) -> str:
         authstring = "{0}:{1}".format(self.username, self.password)
         base64_authorization = base64.b64encode(authstring.encode('utf-8')).decode('utf-8')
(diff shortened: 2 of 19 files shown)


ibm_cloud_sdk_core/authenticators/authenticator.py
```diff
@@ -19,15 +19,15 @@
...
         """Perform the necessary authentication steps for the specified request.
 
         To be implemented by subclasses.
         """
         pass
 
     @abstractmethod
-    def validate(self):
+    def validate(self) -> None:
         """Validates the current set of configuration information in the Authenticator.
```
mkistler, line 32: We should add
```
        Raises:
            ValueError: The configuration information is not valid for service operations.
```

ibm_cloud_sdk_core/authenticators/basic_authenticator.py
```diff
@@ -62,13 +62,13 @@ def validate(self):
                 'Please remove any surrounding {, }, or \" characters.')
 
 
-    def __construct_basic_auth_header(self):
+    def __construct_basic_auth_header(self) -> None:
```
mkistler, line 65: I don't think `None` is correct here.  Three lines down you are returning a string.

ibm_cloud_sdk_core/authenticators/authenticator.py
```diff
@@ -19,15 +19,15 @@
 class Authenticator(ABC):
     """This interface defines the common methods and constants associated with an Authenticator implementation."""
     @abstractmethod
-    def authenticate(self, req: dict):
+    def authenticate(self, req: dict) -> None:
         """Perform the necessary authentication steps for the specified request.
```
mkistler, line 24: I think you should document here that the `req` parameter will be modified to contain the appropriate authentication information.

ibm_cloud_sdk_core/base_service.py
```diff
@@ -118,10 +119,10 @@ def configure_service(self, service_name: str):
                 bool(config.get('DISABLE_SSL'))
             )
 
-    def _set_user_agent_header(self, user_agent_string=None):
+    def _set_user_agent_header(self, user_agent_string=None) -> None:
```
mkistler, line 122: Missing type annotation on `user_agent_string` parameter.
```suggestion
    def _set_user_agent_header(self, user_agent_string: str = None) -> None:
```

I'm sure why `user_agent_string` has a default value -- that seems odd.  I left it there but you might consider removing it.

ibm_cloud_sdk_core/authenticators/iam_authenticator.py
```diff
@@ -56,18 +56,20 @@ class IAMAuthenticator(Authenticator):
 
     def __init__(self,
                  apikey: str,
+                 *,
                  url: Optional[str] = None,
                  client_id: Optional[str] = None,
                  client_secret: Optional[str] = None,
                  disable_ssl_verification: Optional[bool] = False,
```
mkistler, line 63: I don't think we want `Optional` here.

`Optional` should only be used when the default value is `None`.

https://docs.python.org/3/library/typing.html#typing.Optional

ibm_cloud_sdk_core/iam_token_manager.py
```diff
@@ -63,20 +63,21 @@ class IAMTokenManager(JWTTokenManager):
 
     def __init__(self,
                  apikey: str,
+                 *,
                  url: Optional[str] = None,
                  client_id: Optional[str] = None,
                  client_secret: Optional[str] = None,
                  disable_ssl_verification: Optional[str] = False,
```
mkistler, line 70: I don't think we want `Optional` here.

test/test_base_service.py
```diff
@@ -38,19 +39,19 @@ class AnyServiceV1(BaseService):
...
             self,
-            version,
-            service_url=default_url,
-            authenticator=None,
-            disable_ssl_verification=False
-        ):
+            version: str,
+            service_url: Optional[str] = default_url,
+            authenticator: Optional[Authenticator] = None,
+            disable_ssl_verification: Optional[bool] = False
```
mkistler, line 45: I don't think we want `Optional` here.

test/test_base_service.py
```diff
@@ -38,19 +39,19 @@ class AnyServiceV1(BaseService):
 
     def __init__(
             self,
-            version,
-            service_url=default_url,
-            authenticator=None,
-            disable_ssl_verification=False
-        ):
+            version: str,
+            service_url: Optional[str] = default_url,
```
mkistler, line 43: I think this is an incorrect use of `Optional` here.  I think `Optional` should only be used when the default value is `None`.
```suggestion
            service_url: str = default_url,
```

https://docs.python.org/3/library/typing.html

> Optional[t] is added for function and method annotations if a default value equal to `None` is set.

PR #68, opened 2020-04-06
Merge Python3 breaking changes from V2 to master

rmkeezer:
And ensure that the commit message is semantic release friendly for a breaking change 👍


padamstx:
Before we can merge the v2 branch into master, we'll need to do these things:
1. Do a rebase on the current master branch since this PR thinks the v2 branch is out of date.
2. Make sure that each of the commit messages in the v2 branch are correct.  i.e. whichever commits represent breaking changes should have the appropriate "BREAKING CHANGE" mesage.

For #2, this might be all of the commits, in which case we could do this:
1. create a new "v2" branch, maybe call it "v2-new".
2. one at a time, and in the same order as the commits appear in the "v2" branch, cherry-pick a commit from "v2" into "v2-new", then run `git commit --amend` to modify the commit message so that it properly reflects a breaking change.
3. when each commit from "v2" has been handled in step 2, you could remove your local "v2" branch, then rename "v2-new" to be "v2", then run `git push -f` to update in place the "v2" branch (i.e. your PR).
4. after the build is clean, etc. the updated "v2" branch can be merged into master.

PR #72, opened 2020-05-22
refactor: 💡 Refactor JWTTokenManager to introduce TokenManager

bessbd:
Currently, JWTTokenManager is the top-most token manager class which
does not allow introducing new token managers in code that uses the IBM
Python SDK Core.

This commit refactors JWTTokenManager into an abstract
base class of TokenManager and the JWT-specific parts to allow extending
the Python SDK Core with non-JWT token managers.

This PR is analogous to https://github.com/IBM/node-sdk-core/pull/89

### DIFF
--- a/ibm_cloud_sdk_core/jwt_token_manager.py
+++ b/ibm_cloud_sdk_core/jwt_token_manager.py
@@ -1,6 +1,6 @@
 # coding: utf-8
 
-# Copyright 2019 IBM All Rights Reserved.
+# Copyright 2019, 2020 IBM All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -14,16 +14,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import time
-from threading import Lock
+from abc import ABC
 from typing import Optional
 
 import jwt
 import requests
 from .api_exception import ApiException
+from .token_manager import TokenManager
 
 
-class JWTTokenManager:
+# pylint: disable=too-many-instance-attributes
+
+
+class JWTTokenManager(TokenManager, ABC):
     """An abstract class to contain functionality for parsing, storing, and requesting JWT tokens.
 
     get_token will retrieve a new token from the url in case the that there is no existing token,
@@ -44,151 +47,28 @@ class JWTTokenManager:
         the server's SSL certificate should be disabled or not.
         token_name (str): The key used of the token in the dict returned from request_token.
         token_info (dict): The most token_response from request_token.
-        expire_time (int): The time in epoch seconds when the current token within token_info will expire.
-        refresh_time (int): The time in epoch seconds when the current token within token_info should be refreshed.
-        request_time (int): The time the last outstanding token request was issued
-        lock (Lock): Lock variable to serialize access to refresh/request times
-        http_config (dict): A dictionary containing values that control the timeout, proxies, and etc of HTTP requests.
     """
 
-    # pylint: disable=too-many-instance-attributes
-
-    def __init__(self, url: str, *, disable_ssl_verification: bool = False, token_name: Optional[str] = None) -> None:
-        self.url = url
-        self.disable_ssl_verification = disable_ssl_verification
+    def __init__(self, url: str, *, disable_ssl_verification: bool = False, token_name: Optional[str] = None):
+        super().__init__(url, disable_ssl_verification=disable_ssl_verification)
         self.token_name = token_name
         self.token_info = {}
-        self.expire_time = 0
-        self.refresh_time = 0
-        self.request_time = 0
-        self.lock = Lock()
-        self.http_config = {}
-
-    def get_token(self) -> str:
-        """Get a token to be used for authentication.
-
-        The source of the token is determined by the following logic:
-        1.  a) If the current token is expired (or never fetched), make a request for one
-            b) If the curent token should be refreshed, issue a refresh request
-        2. After any requests initiated above complete, return the stored token
-
-        Returns:
-            str: A valid access token
-        """
-        if self._is_token_expired():
-            self.paced_request_token()
-
-        if self._token_needs_refresh():
-            token_response = self.request_token()
-            self._save_token_info(token_response)
-
-        return self.token_info.get(self.token_name)
-
-    def set_disable_ssl_verification(self, status: bool = False) -> None:
-        """Sets the ssl verification to enabled or disabled.
-
-        Args:
-            status: the flag to be used for determining status.
-        """
-        self.disable_ssl_verification = status
-
-    def paced_request_token(self) -> None:
-        """
-        Paces requests to request_token.
-
-        This method pseudo-serializes requests for an access_token
-        when the current token is expired (or has never been fetched).
-        The first caller into this method records its `request_time` and
-        then issues the token request. Subsequent callers will check the
-        `request_time` to see if a request is active (has been issued within
-        the past 60 seconds), and if so will sleep for a short time interval
-        (currently 0.5 seconds) before checking again. The check for an active
-        request and update of `request_time` are serailized by the `lock`
-        variable so that only one caller can become the active requester
-        with a 60 second interval.
-
-        Threads that sleep waiting for the active request to complete will
-        eventually find a newly valid token and return, or 60 seconds will
-        elapse and a new thread will assume the role of the active request.
-        """
-        while self._is_token_expired():
-            current_time = self._get_current_time()
-
-            with self.lock:
-                request_active = self.request_time > (current_time - 60)
-                if not request_active:
-                    self.request_time = current_time
-
-            if not request_active:
-                token_response = self.request_token()
-                self._save_token_info(token_response)
-                self.request_time = 0
-                return
-
-            time.sleep(0.5)  # Sleep for 0.5 seconds before checking token again
-
-
-    def request_token(self) -> None:
-        """Should be overridden by child classes.
-
-        Raises:
-            NotImplementedError: Thrown when called.
-        """
-        raise NotImplementedError(
-            'request_token MUST be overridden by a subclass of JWTTokenManager.'
-        )
-
-    @staticmethod
-    def _get_current_time() -> int:
-        return int(time.time())
-
-    def _is_token_expired(self) -> bool:
-        """
-        Check if currently stored token is expired.
-
-        Returns
-        -------
-        bool
-            True if token is expired; False otherwise
-        """
-        current_time = self._get_current_time()
-        return self.expire_time < current_time
-
-    def _token_needs_refresh(self) -> bool:
-        """
-        Check if currently stored token needs refresh.
-
-        Returns
-        -------
-        bool
-            True if token needs refresh; False otherwise
-        """
-        current_time = self._get_current_time()
-
-        with self.lock:
-            needs_refresh = self.refresh_time < current_time
-            if needs_refresh:
-                self.refresh_time = current_time + 60
-
-        return needs_refresh
 
     def _save_token_info(self, token_response: dict) -> None:
         """
         Decode the access token and save the response from the JWT service to the object's state
-
         Refresh time is set to approximately 80% of the token's TTL to ensure that
         the token refresh completes before the current token expires.
-
         Parameters
         ----------
         token_response : dict
             Response from token service
         """
         self.token_info = token_response
-        access_token = token_response.get(self.token_name)
+        self.access_token = token_response.get(self.token_name)
 
         # The time of expiration is found by decoding the JWT access token
-        decoded_response = jwt.decode(access_token, verify=False)
+        decoded_response = jwt.decode(self.access_token, verify=False)
         # exp is the time of expire and iat is the time of token retrieval
         exp = decoded_response.get('exp')
         iat = decoded_response.get('iat')
--- a/ibm_cloud_sdk_core/token_manager.py
+++ b/ibm_cloud_sdk_core/token_manager.py
@@ -1,0 +1,214 @@
+# coding: utf-8
+
+# Copyright 2020 IBM All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from abc import ABC, abstractmethod
+from threading import Lock
+
+import time
+import requests
+
+from .api_exception import ApiException
+
+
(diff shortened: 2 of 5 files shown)


ibm_cloud_sdk_core/token_manager.py
```diff
@@ -0,0 +1,246 @@
...
+    Attributes:
+        url (str): The url to request tokens from.
+        disable_ssl_verification (bool): A flag that indicates whether verification of
+        the server's SSL certificate should be disabled or not.
+        token_name (str): The key used of the token in the dict returned from request_token.
+        token_info (dict): The most token_response from request_token.
+        expire_time (int): The time in epoch seconds when the current token within token_info will expire.
+        refresh_time (int): The time in epoch seconds when the current token within token_info should be refreshed.
+        request_time (int): The time the last outstanding token request was issued
+        lock (Lock): Lock variable to serialize access to refresh/request times
```
mkistler, line 53: Is there a reason we expose `lock` as an attribute?  Seems like this should be just internal state.

ibm_cloud_sdk_core/token_manager.py
```diff
@@ -0,0 +1,246 @@
...
+        disable_ssl_verification: A flag that indicates whether verification of
+            the server's SSL certificate should be disabled or not. Defaults to False.
+        token_name: The key that maps to the token in the dictionary returned from request_token. Defaults to None.
+
+    Attributes:
+        url (str): The url to request tokens from.
+        disable_ssl_verification (bool): A flag that indicates whether verification of
+        the server's SSL certificate should be disabled or not.
+        token_name (str): The key used of the token in the dict returned from request_token.
+        token_info (dict): The most token_response from request_token.
```
mkistler, line 49: As was done in https://github.com/IBM/node-sdk-core/pull/89, I'd like you simplify the TokenManager class to only hold the `access_token`, leaving the details of `token_name` and `token_info` to implementing subclasses.

Following the Node implementation, that will mean making `save_token_info` an abstract method that subclasses implement that must set `access_token`, `expire_time`, and `refresh_time`.

This will simplify the implementation and make it more consistent with the Node core.

ibm_cloud_sdk_core/jwt_token_manager.py
```diff
@@ -13,17 +13,14 @@
...
 
 import jwt
-import requests
-from .api_exception import ApiException
+
+from .token_manager import TokenManager
 
 
-class JWTTokenManager:
+class JWTTokenManager(TokenManager, ABC):
```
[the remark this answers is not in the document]
bessbd: Good call. I forgot to add the [abstractmethod](https://docs.python.org/3/library/abc.html#abc.abstractmethod) decorators to the abstract methods. Done in c27ded1
Making the token managers real [abstract base class](https://docs.python.org/3/library/abc.html)es makes them non-instantiable which is desired as they are not to be used directly, but to be extended in subclasses such as [CP4DTokenManager](https://github.com/IBM/python-sdk-core/blob/master/ibm_cloud_sdk_core/cp4d_token_manager.py).

ibm_cloud_sdk_core/jwt_token_manager.py
```diff
@@ -206,21 +67,15 @@ def _request(self,
                  data=None,
                  auth_tuple=None,
                  **kwargs) -> dict:
-        kwargs = dict({"timeout": 60}, **kwargs)
-        kwargs = dict(kwargs, **self.http_config)
-
-        if self.disable_ssl_verification:
-            kwargs['verify'] = False
+        return self._request_raw(method, url, headers=headers, params=params, data=data,
+                                 auth_tuple=auth_tuple, **kwargs).json()
```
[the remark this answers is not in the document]
bessbd: [_request](https://github.com/IBM/python-sdk-core/pull/72/files/146b4dc7f5adf7a612173ce4d41ed292035dae49#diff-4073e5945d5ebc449fcfce455620beccR61) was doing two things: making a prepared request via `requests.request` and a [json deserialization](https://github.com/IBM/python-sdk-core/pull/72/files/146b4dc7f5adf7a612173ce4d41ed292035dae49#diff-4073e5945d5ebc449fcfce455620beccL224). This deserialization hides information necessary to some alternative token managers - for example one that needs the cookies from the response so I've extracted [_request_raw](https://github.com/IBM/python-sdk-core/pull/72/files/146b4dc7f5adf7a612173ce4d41ed292035dae49#diff-5745e4f083a8d956453c9fe4e33e69bbR205) that only performs the request and exception throwing (so that is not lost, just moved to `TokenManager`).

ibm_cloud_sdk_core/token_manager.py
bessbd: It was already exposed in [JWTTokenManager](https://github.com/IBM/python-sdk-core/blob/master/ibm_cloud_sdk_core/jwt_token_manager.py#L50). That allows "external" code to utilize the lock for synchronization. That said I'm okay with moving it to internal. Let me know if you'd like me to make it "internal use".

ibm_cloud_sdk_core/token_manager.py
bessbd: Done in 07136fd

ibm_cloud_sdk_core/token_manager.py
```diff
@@ -0,0 +1,246 @@
...
+from abc import ABC
+from threading import Lock
+from typing import Optional
+import time
+import requests
+
+from .api_exception import ApiException
+
+
+class TokenManager(ABC):
```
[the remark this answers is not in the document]
bessbd: Same reason: https://github.com/IBM/python-sdk-core/pull/72#discussion_r429796347

bessbd:
Thank you for the review, @mkistler !
I think I've addressed all the comments you've added. Please let me know what you think.

ibm_cloud_sdk_core/jwt_token_manager.py
```diff
@@ -51,151 +48,15 @@ class JWTTokenManager:
...
-        Refresh time is set to approximately 80% of the token's TTL to ensure that
-        the token refresh completes before the current token expires.
-
-        Parameters
-        ----------
-        token_response : dict
-            Response from token service
-        """
-        self.token_info = token_response
+    def extract_exp_and_ttl(self, token_response):
```
vmatyusGitHub: `extract_exp_and_ttl` function also makes token parsing and storing, so the name of the function does not fully cover its full responsibility.
I would vote to have the `_save_token_info` again as abstract function that should be implement with tokenManager specific methods.

ibm_cloud_sdk_core/jwt_token_manager.py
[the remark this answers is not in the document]
vmatyusGitHub: Please also take consider you can change the content of `request_token` functions in `IAMTokenManager` and in `CP4DTokenManager` to have `return response.json()` there.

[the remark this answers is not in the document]
vmatyusGitHub: Or an other solution aspect `_request` would be good to keep in `TokenManager` level and in `JWTTokenManager` would better to have a `_jwt_request` and this is called from `IAMTokenManager` and `CP4DTokenManager`.

ibm_cloud_sdk_core/jwt_token_manager.py
bessbd: I believe this was done in 07136fd526ef48939bbc44233c3cb8a08eeb1b01 . Let me now if I've missed anything there.

ibm_cloud_sdk_core/jwt_token_manager.py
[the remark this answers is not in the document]
bessbd: > Please also take consider you can change the content of `request_token` functions in `IAMTokenManager` and in `CP4DTokenManager` to have `return response.json()` there.

`IAMTokenManager` and `CP4DTokenManager` (unchanged) use [JWTTokenManager's _request](https://github.com/IBM/python-sdk-core/pull/72/files/146b4dc7f5adf7a612173ce4d41ed292035dae49#diff-4073e5945d5ebc449fcfce455620beccR61) which does `.json()` so we should be okay.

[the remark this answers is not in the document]
bessbd: > Or an other solution aspect `_request` would be good to keep in `TokenManager` level and in `JWTTokenManager` would better to have a `_jwt_request` and this is called from `IAMTokenManager` and `CP4DTokenManager`.

Sorry, I don't understand. What would `_jwt_request` do?

bessbd:
Thank you for the review, @vmatyusGitHub !
I've left a response to your comments in https://github.com/IBM/python-sdk-core/pull/72#discussion_r429923545 . For some reason it doesn't show up here in the conversation tab for me.

ibm_cloud_sdk_core/jwt_token_manager.py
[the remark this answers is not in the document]
vmatyusGitHub: I think it would worth thinking about a function renaming `_request` -> `_jwt_request` and `_request_raw` -> `_request`.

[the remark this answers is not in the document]
bessbd: > `_request_raw` -> `_request`

That sounds like an API-breaking change as subclasses of JWTTokenManager (eg. [IAMTokenManager](https://github.com/IBM/python-sdk-core/blob/master/ibm_cloud_sdk_core/iam_token_manager.py#L109)) use 
 JWTTokenManager's _request and they expect a `.json()`-ed value.
I can do the follow-up this repo, but all external client code that inherit from `python-sdk-core`'s JWTTokenManager would need migration.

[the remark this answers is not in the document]
mkistler: While I like @vmatyusGitHub's suggested renaming, I think we should avoid API-breaking changes, so my preference is to leave the names as is.

ibm_cloud_sdk_core/token_manager.py
mkistler: Good point.  I suppose it would technically be a breaking change to make it internal, and while I think the chance of actually breaking anyone is low, there's more risk than value in making the change so we should leave it as is.

ibm_cloud_sdk_core/token_manager.py
```diff
@@ -0,0 +1,224 @@
...
+            str: A valid access token
+        """
+        if self._is_token_expired():
+            self.paced_request_token()
+
+        if self._token_needs_refresh():
+            token_response = self.request_token()
+            self._save_token_info(token_response)
+
+        return self.extract_token_from_stored_response()
```
mkistler, line 84: I suppose this works but I think I would prefer that the `access_token` be defined as a property that is set by `_save_token_info` -- along with `expire_time` and `refresh_time` -- so that here we can simply have:
```
    return self.access_token
```

ibm_cloud_sdk_core/token_manager.py
```diff
@@ -0,0 +1,224 @@
...
+
+        raise ApiException(response.status_code, http_response=response)
+
+    def _set_expire_and_refresh_time(self, exp, ttl):
+        self.expire_time = exp
+        buffer = ttl * 0.2
+        self.refresh_time = self.expire_time - buffer
+
+    @abstractmethod
+    def extract_token_from_stored_response(self):  # pragma: no cover
```
mkistler, line 218: If we store the token directly then we don't need this method.

ibm_cloud_sdk_core/token_manager.py
```diff
@@ -0,0 +1,224 @@
...
+            params=params,
+            data=data,
+            auth=auth_tuple,
+            **kwargs)
+        if 200 <= response.status_code <= 299:
+            return response
+
+        raise ApiException(response.status_code, http_response=response)
+
+    def _set_expire_and_refresh_time(self, exp, ttl):
```
mkistler, line 212: I'm not sure we need this method.  Subclasses could simply implement this directly in their `_save_token_info` methods.  But if we leave it, it should have a doc string that describes its intended behavior.

ibm_cloud_sdk_core/jwt_token_manager.py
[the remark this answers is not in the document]
mkistler: Actually, as I think about this more, I think it would be my preference to leave the request code here as is and simply leave it to other subclasses to implement their own request method.  Token manager really has no need for its own request method -- it doesn't use it for anything. So let subclasses implement this as they choose.

ibm_cloud_sdk_core/token_manager.py
bessbd: Done in dc4162a

Mind that, this alters semantics slightly: https://github.com/IBM/python-sdk-core/blob/master/ibm_cloud_sdk_core/jwt_token_manager.py#L85 so a minor test change was necessary: https://github.com/IBM/python-sdk-core/pull/72/commits/dc4162adc3fec8d26d91fa0b6addb8b5f3c239e5#diff-6bc545e548b4e369e6823510e07c95ceR65

ibm_cloud_sdk_core/token_manager.py
bessbd: Done in dc4162a

ibm_cloud_sdk_core/token_manager.py
bessbd: Inlined in 949f389

bessbd:
Hi @mkistler ,
I think I've addressed all your review feedback so far. Is there anything else you'd like changed for this PR?

ibm_cloud_sdk_core/jwt_token_manager.py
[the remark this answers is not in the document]
mkistler: I suppose the code is fine as is but let me throw out another alternative:
- implement `_request` in `JWTTokenManager` as it had been -- without calling `_request_raw`.
- Change `_request_raw` to just `_request` in `TokenManager` leaving the rest of the implementation as is.

This does duplicate some functionality, but sets us up for an easy transition in the next major release where we'd switch `JWTTokenManager` to use `TokenManager`'s `_request` and eliminate that duplicate functionality.  So we endure a little duplication now for the sake of compatibility but pave the way for a "clean" implementation later.

I think this suggestion makes the transition easier because subclasses of `TokenManager` can use `_request` now and won't have to change in the next major release.  (I suppose we could just continue to use `_request_raw` in the next major release, but I find that unappealing.)

How does this sound?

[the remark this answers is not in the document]
bessbd: I've just pushed f6ffb41 . Is this what you've meant?

ibm_cloud_sdk_core/jwt_token_manager.py
```diff
@@ -224,3 +104,9 @@ def _request(self,
             return response.json()
 
         raise ApiException(response.status_code, http_response=response)
+
+    @abstractmethod
+    def request_token(self) -> None:  # pragma: no cover
+        """Should be overridden by child classes.
+        """
+        pass
```
mkistler, line 112: Do we need this here, since it is already defined in `TokenManager`?

bessbd: Good catch! I've just removed it.
If you think this PR is okay now, can you please merge it?

bessbd:
🎉!
Thank you for the review, @mkistler and @vmatyusGitHub !
Also, thank you for the merge, @mkistler

PR #80, opened 2020-09-17
feat(IAM Authenticator): add support for optional 'scope' property

jorge-ibm:
The IAM /identity/token (get token) operation supports an optional "scope" form parameter which can be used to obtain an IAM access token having a limited set of scopes in this it can be used.

This commit adds support for the "Scope" field within the IamTokenManager class. If set, the value of this field is sent as the "scope" form param in the "get token" request body when obtaining an access token.

ref: https://github.ibm.com/arf/planning-sdk-squad/issues/2167

### DIFF
--- a/resources/ibm-credentials.env
+++ b/resources/ibm-credentials.env
@@ -10,3 +10,12 @@ SERVICE_1_CLIENT_ID=somefake========id
 SERVICE_1_CLIENT_SECRET===my-client-secret==
 SERVICE_1_AUTH_URL=https://iamhost/iam/api=
 SERVICE_1_URL=service1.com/api
+
+# Service2 configured with IAM w/scope
+SERVICE_2_AUTH_TYPE=iam
+SERVICE_2_APIKEY=V4HXmoUtMjohnsnow=KotN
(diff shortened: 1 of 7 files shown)


resources/ibm-credentials.env
```diff
@@ -9,4 +9,13 @@ SERVICE_1_APIKEY=V4HXmoUtMjohnsnow=KotN
 SERVICE_1_CLIENT_ID=somefake========id
 SERVICE_1_CLIENT_SECRET===my-client-secret==
 SERVICE_1_AUTH_URL=https://iamhost/iam/api=
-SERVICE_1_URL=service1.com/api
\ No newline at end of file
+SERVICE_1_URL=service1.com/api
+
+# Service2 configured with IAM w/scope
+SERVICE_2_AUTH_TYPE=iam
+SERVICE_2_APIKEY=V4HXmoUtMjohnsnow=KotN
```
padamstx, line 16: No matter what @dpopp07 says, I'm ok with you spelling this "john" instead of "jon" :)

padamstx (approved):
LGTM

PR #81, opened 2020-10-06
feat: allow gzip compression on request bodies

rmkeezer:
Allows gzip compression to be enabled for request bodies in the base service. This should be enabled by either the service or by an external config.


ibm_cloud_sdk_core/base_service.py
```diff
@@ -245,6 +249,9 @@ def send(self, request: requests.Request, **kwargs) -> DetailedResponse:
             logging.exception('Error in service call')
             raise
 
+    def enable_gzip_compression(self, should_enable_compression: bool = False) -> None:
```
padamstx, line 254:
```suggestion
    def set_enable_gzip_compression(self, should_enable_compression: bool = False) -> None:
```
And I think it would be good if we also had a getter for this flag as well.

ibm_cloud_sdk_core/base_service.py
```diff
@@ -81,6 +82,7 @@ def __init__(self,
         self.authenticator = authenticator
         self.disable_ssl_verification = disable_ssl_verification
         self.default_headers = None
+        self.gzip_compression = False
```
padamstx, line 87: Why not allow this new flag to be passed in via an optional parameter, similar to disable_ssl_verification?

padamstx: Also need to add "gzip_compression" to the list of attributes in the comments above.

ibm_cloud_sdk_core/base_service.py
```diff
@@ -41,6 +42,8 @@
 # import http.client as http_client
 # http_client.HTTPConnection.debuglevel = 1
 
+#pylint: disable=too-many-instance-attributes
+#pylint: disable=too-many-locals
```
rmkeezer, line 46: We have 8/7 instance attributes and 17/15 locals with the newly added `gzip_compression` variables, so I had to disable these lint checks for this class.

ibm_cloud_sdk_core/base_service.py
```diff
@@ -309,6 +319,17 @@ def prepare_request(self,
 
         self.authenticator.authenticate(request)
 
+        # Compress the request body if applicable
+        current_encoding = request['headers']
```
padamstx, line 323: I got a little confused by the header handling code in this function, until I expanded my view to see where "headers" was initialized above.   My suggestion would be use the local variable "headers" exclusively as the dict containing the current set of request headers, until it's time to simply copy them to `request['headers']` near the end of prepare_request().

ibm_cloud_sdk_core/base_service.py
```diff
@@ -309,6 +319,17 @@ def prepare_request(self,
...
+        if (self.gzip_compression and
+                'content-encoding' not in current_encoding and
```
padamstx, line 325:
```suggestion
                'content-encoding' not in headers and
```

padamstx (changes requested):
I left a few comments that need to be addressed, plus I think we need to add a test or two related to the ENABLE_GZIP config property.

ibm_cloud_sdk_core/base_service.py
rmkeezer: Good point, done

ibm_cloud_sdk_core/base_service.py
rmkeezer: Right, I'm not sure why I needed to make a separate local variable for this, so I'm just the existing headers now

jorge-ibm (commented):
Looks good. Could we maybe add 1 more test to test non-json request bodies to make sure that's compressed correctly too ?

rmkeezer:
I added another test for reading the ENABLE_GZIP config and using non-json data

jorge-ibm (approved):

padamstx (approved):
LGTM

PR #84, opened 2020-11-04
feat(BaseService): Use a `requests.Session` object to configure retries and other configuration

### LINKED ISSUE #83: Implementing retries in `BaseService`
`BaseService.prepare_request()` and `BaseService.send()` perform `requests` directly, rather than through a `requests.Session()` object, which makes it difficult to add a Transport Adapter to implement exponential backoff.

It is possible to implement retries through abuse of the `http_config` dict, using request hooks to perform recursive calls, but this solution is worse than wrapping a request.
Request wrapping is undesirable because the implementation will inevitably be worse than the one implemented in `urllib3`, and introduce noise and boilerplate into every request written.

Similarly to [this blogpost](https://findwork.dev/blog/advanced-usage-python-requests-timeouts-retries-hooks/#retry-on-failure), if a `requests.Session()` object is used to perform all requests, this allows users to simply mount a Transport Adapter, which will allow them to easily implement global retries and timeouts according to their needs.

I've written a PR for the change, which should be very low impact and backwards-compatible.

illeatmyhat:
Closes #83

Summary: Using a `requests.Session` object allows the client to apply flexible global configuration, i.e. exponential backoff retries, timeouts, proxies, etc. to every request using a Transport Adapter such as `HTTPAdapter`.

Usage:
```python
retry_strategy = Retry(
    total=3,
    backoff_factor=1,
    status_forcelist=[429, 500, 502, 503, 504],
    method_whitelist=["HEAD", "GET", "OPTIONS", "POST"])
adapter = HTTPAdapter(max_retries=retry_strategy)
service.http_client.mount('https://', adapter)
detailed_response = service.send(prepped)
```
An implementer service could also apply this in their `__init__()` function.

Unfortunately I could not introduce a new test case, because the `responses` library does not support testing retries.
https://github.com/getsentry/responses/issues/135

### DIFF
--- a/ibm_cloud_sdk_core/authenticators/iam_authenticator.py
+++ b/ibm_cloud_sdk_core/authenticators/iam_authenticator.py
@@ -45,6 +45,8 @@ class IAMAuthenticator(Authenticator):
         proxies: Dictionary for mapping request protocol to proxy URL. Defaults to None.
         proxies.http (optional): The proxy endpoint to use for HTTP requests.
         proxies.https (optional): The proxy endpoint to use for HTTPS requests.
+        scope: The "scope" to use when fetching the bearer token from the IAM token server.
+        This can be used to obtain an access token with a specific scope.
 
     Attributes:
         token_manager (IAMTokenManager): Retrives and manages IAM tokens from the endpoint specified by the url.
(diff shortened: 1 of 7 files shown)


padamstx:
@illeatmyhat Thank you for opening this PR.  We'll try to get this reviewed and get you some feedback soon.

rmkeezer:
I botched this PR when trying to update it with master, I remade the PR here: https://github.com/IBM/python-sdk-core/pull/101 using your commit on my fork

PR #85, opened 2020-11-06
fix: replace zlib with gzip for gzip compression

rmkeezer:
ref: https://github.ibm.com/arf/planning-sdk-squad/issues/2290

Using gzip instead of zlib will avoid an issue when compressing and decompressing gzip file data.


padamstx (approved):
Changes look good.
I don't remember if you already modified the python generator to emit some gzip-related code as part of the unit tests or not.  But if not, we should definitely do that.   You could perhaps take a look at what I did with the Go generator when I added gzip support for inspiration.

PR #86, opened 2020-11-16
fix: use millisecond precision when marshalling date-time values

padamstx:
Fixes arf/planning-sdk-squad#2313

This commit makes a minor change to the datetime_to_string function
so that it uses millisecond precision when converting a datetime value
to a string.  This is aligned with the API handbook requiremetns re:
the wire format for date-time values.
I also added a bunch of tests that are (more or less) similar to
the set of tests in the Java and Go cores.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,7 +48,7 @@ coverage.xml
 .vscode
 
 # virtual env
-venv/
+venv*/
 # python 3 virtual env
 python3/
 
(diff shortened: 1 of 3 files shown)


padamstx:
Looks like all 4 of the builds failed due to me using a feature that is not available in any of the python versions that we currently support.   THANKS PYTHON!!!
I'll regroup and update this PR.

padamstx:
Closing this and will open a new PR.

PR #87, opened 2020-11-16
test: add tests re: date-time marshalling/unmarshalling

padamstx:
Fixes arf/planning-sdk-squad#2313

This commit adds new tests for the datetime_to_string and
string_to_datetime functions to ensure that each of the required
date-time formats are supported.  These new tests are aligned
with similar tests in the Java and Go cores.

Edit: I originally included a change to datetime_to_string() to use only milliseconds precision when marshalling a datetime value, but it turns out that the python versions we build with on Travis do not support the use of the "timespec" parameter on the datetime.isoformat() function.   So, I'll leave datetime_to_string() alone for now and it will continue to produce strings with microsecond precision, which leaves us no worse off than we were before.   But the addition of the new tests should provide us with some confidence that we support the same formats of datetime values as in the Java and Go cores.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,7 +48,7 @@ coverage.xml
 .vscode
 
 # virtual env
-venv/
+venv*/
 # python 3 virtual env
 python3/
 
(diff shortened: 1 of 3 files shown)


mkistler (approved):
Looks good! 👍

rmkeezer (approved):
LGTM

PR #89, opened 2021-01-04
fix: use options dict for jwt.decode verify flag

### LINKED ISSUE #88: Decoding access_token does not work with PyJWT v2.0.0
The requirement states `PyJWT >=1.7.1`, since PyJWT v2.0.0 was released, on a fresh install v2.0.0 is installed which throws the below error
image

psschwei:
Closes  #88

The `verify` parameter for the `jwt.decode` function has been deprecated as of PyJWT v2.0.0. It has been moved into an `options` dictionary with key `verify_signature`. This PR updates the code to use the new dict.

Note that when I cloned the repo there were several failing tests unrelated to this code change (the same tests failed both before and after the code change).

### DIFF
--- a/ibm_cloud_sdk_core/jwt_token_manager.py
+++ b/ibm_cloud_sdk_core/jwt_token_manager.py
@@ -68,7 +68,7 @@ class JWTTokenManager(TokenManager, ABC):
         self.access_token = token_response.get(self.token_name)
 
         # The time of expiration is found by decoding the JWT access token
-        decoded_response = jwt.decode(self.access_token, verify=False)
+        decoded_response = jwt.decode(self.access_token, options={"verify_signature": False})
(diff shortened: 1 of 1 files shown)


rmkeezer:
Thanks for PR! In addition to the changes to the verify signature, it seems pyjwt also needs algorithms to be specified and a few other changes, I opened a PR here https://github.com/IBM/python-sdk-core/pull/92 including both changes. We have a new patch release that limits pyjwt to <2.x until these changes are good to go.

padamstx:
@psschwei Thank you for submitting this PR, but we had a bit of a firedrill over the last couple of days in this area and the net result is that we've incorporated your change into a slightly larger PR (https://github.com/IBM/python-sdk-core/pull/91).   I'll close this one out.   Please let us know if you run into any other problems with the python core.   Thanks!

PR #90, opened 2021-01-05
chore: update PyJWT version to <2.0.0 and restrict other req versions

rmkeezer:
PyJWT version >=2.0.0 had a recent breaking change, this restricts the version until we update our code to use the new change. This also adds a max version to the other requirements


padamstx (approved):
LGTM

PR #92, opened 2021-01-06
fix: update pyjwt parameters to satisfy 2.x changes

rmkeezer:
This change includes updates to the jwt.decode call from by updating the verify options from https://github.com/IBM/python-sdk-core/pull/89 and specifying the algorithm in the decode to satisfy new changes in version 2.0.0 of pyjwt.

### DIFF
--- a/test/test_base_service.py
+++ b/test/test_base_service.py
@@ -111,7 +111,7 @@ def get_access_token() -> str:
         headers={
             'kid': '230498151c214b788dd97f22b85410a5'
         })
-    return access_token.decode('utf-8')
+    return access_token
 
 
 @responses.activate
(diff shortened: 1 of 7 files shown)


test/test_base_service.py
```diff
@@ -111,7 +111,7 @@ def get_access_token() -> str:
         headers={
             'kid': '230498151c214b788dd97f22b85410a5'
         })
-    return access_token.decode('utf-8')
+    return access_token
```
rmkeezer, line 114: In the new version pyjwt encode now returns an encoded string instead of bytes so decode is not necessary here anymore

padamstx (approved):
LGTM

PR #93, opened 2021-01-08
fix: update setup.py requirements to match requirements.txt

rmkeezer:
The setup.py requirements should match the requirements.txt. Using the same method from the python sdk template, this change loads setup the reqs from the requirements.txt file


padamstx (approved):
LGTM

PR #94, opened 2021-01-08
fix: include requirements.txt in MANIFEST.in for setup.py reference

rmkeezer:
requirements.txt needs to be included in the MANIFEST.in file to be included in setup.py


padamstx (approved):
LGTM

PR #101, opened 2021-01-14
feat(BaseService): Use a `requests.Session` object to configure retries and other configuration

### LINKED ISSUE #83: Implementing retries in `BaseService`

rmkeezer:
Closes #83

Updated from https://github.com/IBM/python-sdk-core/pull/84

### DIFF
--- a/ibm_cloud_sdk_core/base_service.py
+++ b/ibm_cloud_sdk_core/base_service.py
@@ -65,6 +65,8 @@ class BaseService:
         default_headers (dict): A dictionary of headers to be sent with every HTTP request to the service endpoint.
         jar (http.cookiejar.CookieJar): Stores cookies received from the service.
         http_config (dict): A dictionary containing values that control the timeout, proxies, and etc of HTTP requests.
+        http_client (Session): A configurable session which can use Transport Adapters to configure retries, timeouts,
+            proxies, etc. globally for all requests.
         enable_gzip_compression (bool): A flag that indicates whether to enable gzip compression on request bodies
     Raises:
         ValueError: If Authenticator is not provided or invalid type.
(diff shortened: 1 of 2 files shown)


padamstx (commented):
A few questions before we call this one done:
1. Should we provide a convenient way for the user to set their own http client (Session) object?   Perhaps a BaseService.set_http_client() method, and if we provide a setter we should probably also provide a getter to make it easy to retrieve the client and alter its configuration.
2. How is the `disable_ssl_verification` feature affected if the user supplies his own client (Session) object?
3. Given that we will now be (re-)using an exlicitly created client (aka Session) object, should we also provide a convenient way to enable retries, similar to what we do in Go?   That would imply that we add a BaseService.enable_retries() method of some sort.

rmkeezer:
1. I added a get_http_client and set_http_client method but you can also set the base_service.http_client directly if necessary
2. It shouldn't affect `disable_ssl_verification` since the flag is passed into the request kwargs during the `prepare_request()` function so even if the user sets their own http_client, the flag will be passed into that session during a request
3. That's a good idea and I opened an issue here to add retry logic to the core here: https://github.ibm.com/arf/planning-sdk-squad/issues/2444 I believe this should not be difficult for python since the requests library and the urllib3 library do a lot of heavy lifting with retry logic configuration, the option just needs to be added to pass down config parameters.

padamstx (approved):
LGTM

PR #103, opened 2021-01-28
feat(python): add max retry configuration for python requests session

rmkeezer:
Adds retry logic configuration for the python requests http_client. `MAX_RETRIES` determine how many times the client will retry the request and `RETRY_INTERVAL` will determine how long between each retry.

For the Python requests session, the `RETRY_INTERVAL` is passed to the `backoff_factor` instead of being "seconds between retries" since there is no retry interval. The `backoff_factor` value is passed to a function documented [here](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#module-urllib3.util.retry). Ex. a `backoff_factor` of 0.1 will result in retry intervals of [0.0s, 0.2s, 0.4s, 0.8s, 1.6s, etc]


ibm_cloud_sdk_core/base_service.py
```diff
@@ -136,6 +159,14 @@ def configure_service(self, service_name: str) -> None:
             )
         if config.get('ENABLE_GZIP') is not None:
             self.set_enable_gzip_compression(config.get('ENABLE_GZIP') == 'True')
+        if config.get('MAX_RETRIES'):
```
padamstx, line 162: I think we should support the same set of service-level config properties that we support in the Go core:
ENABLE_RETRIES (boolean)
MAX_RETRIES
RETRY_INTERVAL
where ENABLE_RETRIES must be set to true for retries to be enabled, in which case we look at the values for the other two properties and use those for configuration if present (otherwise we use default values).
So to be clear, if ENABLE_RETRIES is set to true and is the only property set, then retries are enabled with defaults for max retries and the interval.

ibm_cloud_sdk_core/base_service.py
```diff
@@ -82,9 +84,14 @@ def __init__(self,
                  service_url: str = None,
                  authenticator: Authenticator = None,
                  disable_ssl_verification: bool = False,
-                 enable_gzip_compression: bool = False) -> None:
+                 enable_gzip_compression: bool = False,
+                 max_retries: int = 10,
```
padamstx, line 88: I'm leaning toward the notion that the max_retries and retry_interval parameters should not be included in the ctor.  Instead, let's introduce two methods:
1. `enable_retries(max_retries, retry_interval)`
2. `disable_retries()`
This is simlar to the way in which the Go core works.

ibm_cloud_sdk_core/base_service.py
```diff
@@ -98,6 +105,22 @@ def __init__(self,
             raise ValueError(
                 'authenticator should be of type Authenticator')
 
+    def setup_http_client(self) -> None:
```
padamstx, line 108: This would effectively become the `enable_retries(max_retries, retry_interval)` method.

ibm_cloud_sdk_core/base_service.py
```diff
@@ -136,6 +159,14 @@ def configure_service(self, service_name: str) -> None:
             )
         if config.get('ENABLE_GZIP') is not None:
             self.set_enable_gzip_compression(config.get('ENABLE_GZIP') == 'True')
+        if config.get('MAX_RETRIES'):
+            self.set_max_retries(
+                int(config.get('MAX_RETRIES'))
```
padamstx, line 164: With the new `enable_retries(...)` method, if the ENABLE_RETRIES property is set to true, then we would retrieve the MAX_RETRIES and RETRY_INTERVAL properties, then call `enable_retries()` instead of having individual setters for max_retries and retry_interval.

ibm_cloud_sdk_core/base_service.py
```diff
@@ -168,6 +199,26 @@ def set_disable_ssl_verification(self, status: bool = False) -> None:
         """
         self.disable_ssl_verification = status
 
+    def set_max_retries(self, max_retries: int = 10) -> None:
```
padamstx, line 202: These methods go away after adding `enable_retries()`.

resources/ibm-credentials-retry.env
```diff
@@ -0,0 +1,5 @@
+INCLUDE_EXTERNAL_CONFIG_APIKEY=mockkey
+INCLUDE_EXTERNAL_CONFIG_AUTH_TYPE=iam
+INCLUDE_EXTERNAL_CONFIG_URL=https://mockurl
+INCLUDE_EXTERNAL_CONFIG_MAX_RETRIES=5
```
padamstx, line 4: Add `INCLUDE_EXTERNAL_CONFIG_ENABLE_RETRIES=true`

test/test_base_service.py
```diff
@@ -553,6 +553,27 @@ def test_gzip_compression_external():
     assert prepped['data'] == gzip.compress(b'{"foo": "bar"}')
     assert prepped['headers'].get('content-encoding') == 'gzip'
 
+@responses.activate
+def test_retry_config_external():
+    # Should set gzip compression from external config
```
padamstx, line 558: I don't think this comment applies.  Copy/paste error?

test/test_base_service.py
```diff
@@ -553,6 +553,27 @@ def test_gzip_compression_external():
     assert prepped['data'] == gzip.compress(b'{"foo": "bar"}')
     assert prepped['headers'].get('content-encoding') == 'gzip'
 
+@responses.activate
+def test_retry_config_external():
```
padamstx, line 557: I think we need several different tests to put the retries support through its paces.  For examples of the types of tests you could look at `base_service_retries_test.go` in the Go core.

padamstx (changes requested):
I commented on a number of things that I'd like to see changed.  My perspective is that we should try to align as closely with the Go core behavior as much as we can in terms how retries are enabled, configured via external config properties, etc.

I think we need a much larger set of tests in the python core for retries.  Please see the retries-related tests in the Go core for ideas on things to test.

Also, in conjunction with this PR I would expect that there will be changes to the Python unit test generation that will exercise retries in the unit tests, similar to what was done for Go.

Happy to discuss any of this with you in more detail.

padamstx (approved):
LGTM

mkistler (approved):
Looks good! 👍

Issue #60, opened 2020-03-19
Remove custom messaging in base_service.py


ajaykumar7-ibm:
The custom exception message in base_service.py does not give much control to the users when handling an exception at their side. It is a bit annoying for our customers too.

Though we are handling this exception our side, our customer is seeing exception logging msgs logged by base_service.py at lines
https://github.com/IBM/python-sdk-core/blob/master/ibm_cloud_sdk_core/base_service.py#L227
https://github.com/IBM/python-sdk-core/blob/master/ibm_cloud_sdk_core/base_service.py#L230
https://github.com/IBM/python-sdk-core/blob/master/ibm_cloud_sdk_core/base_service.py#L233

Is it possible to remove these custom messages so that customers whoever is using this library, will be able to handle and log exception the way they want?

For example:
We are getting this msg logged.
```
ERROR:root:Unknown error
Traceback (most recent call last):
  File "/Users/env/lib/python3.8/site-packages/ibm_cloud_sdk_core/base_service.py", line 224, in send
    raise ApiException(
ibm_cloud_sdk_core.api_exception.ApiException: Error: Unknown error, Code: 409

```
the response which we wants to show to our customer-
```
{'detail': 'Document already exists: xyz', 'instance': 'xyz', 'status': 409, 'title': 'Conflict', 'type': 'about:blank'}
```

mkistler:
Maybe similar to the problem fixed in https://github.com/IBM/python-sdk-core/pull/59

rmkeezer:
As pointed out, it seems that changes from https://github.com/IBM/python-sdk-core/pull/59 allows custom messages for non-401 errors so you should be able to specify the 409 error with your custom message now. @ajaykumar7-ibm does this solve your with the latest changes?

gilv:
@rmkeezer do you plan to address this issue? We use latest sdk and suffer from those 500 error prints which are very annoying. I guess #59 didn't resolved it.. Please see this comment:

ibm_cloud_sdk_core lib always prints this minimum exception:

```
Traceback (most recent call last):
  File "/home/lachlan/miniconda3/envs/sm38/lib/python3.8/site-packages/ibm_cloud_sdk_core/base_service.py", line 246, in send
    raise ApiException(
ibm_cloud_sdk_core.api_exception.ApiException: Error: Internal server error, Code: 500

```

This is caused by this line. I don't know why, but they use the base logging lib to print the exception (in root logging) and then they call raise

```
except ApiException as err:
[...]

PR #104, opened 2021-02-01
fix: silently ignore missing configuration file

padamstx:
While attempting to read configuration properties from
the config file specified via ${IBM_CREDENTIALS_FILE}, the
read_external_sources() function should silently ignore
an error if the file doesn't exist, and then proceed
to the next step of the config algorithm.
This was not working correctly, but this commit fixes that.

Also fixes: #60


rmkeezer (approved):
LGTM!

PR #108, opened 2021-02-25
feat: expose refresh token in iam authenticator

rmkeezer:
The IAM Authenticator now exposes the `refresh_token` variable which is set when a `get_token()` request is made and a refresh token is returned.

### DIFF
--- a/ibm_cloud_sdk_core/iam_token_manager.py
+++ b/ibm_cloud_sdk_core/iam_token_manager.py
@@ -60,7 +60,7 @@ class IAMTokenManager(JWTTokenManager):
         scope: The "scope" to use when fetching the bearer token from the IAM token server.
         This can be used to obtain an access token with a specific scope.
     """
-    DEFAULT_IAM_URL = 'https://iam.cloud.ibm.com/identity/token'
+    DEFAULT_IAM_URL = 'https://iam.cloud.ibm.com'
     CONTENT_TYPE = 'application/x-www-form-urlencoded'
     OPERATION_PATH = "/identity/token"
     REQUEST_TOKEN_GRANT_TYPE = 'urn:ibm:params:oauth:grant-type:apikey'
--- a/ibm_cloud_sdk_core/iam_token_manager.py
+++ b/ibm_cloud_sdk_core/iam_token_manager.py
@@ -17,6 +17,7 @@
 from typing import Dict, Optional
 from .jwt_token_manager import JWTTokenManager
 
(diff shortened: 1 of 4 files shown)


padamstx (commented):
The PR looks good.  However, I think we should take this opportunity to introduce an integration test that involves the actual IAM token service.

test/test_iam_token_manager.py
```diff
@@ -256,3 +258,86 @@ def test_request_token_auth_in_setter_scope():
...
+# IAMTEST2_AUTH_URL=<url> e.g. https://iam.test.cloud.ibm.com/identity/token
+# IAMTEST2_AUTH_TYPE=iam
+# IAMTEST2_APIKEY=<apikey>
+# IAMTEST2_CLIENT_ID=<client id>
+# IAMTEST2_CLIENT_SECRET=<client secret>
+
+# Then uncomment the function below and run this command:
+# python3 -m pytest test -k "test_iam_live_token_server"
+
+# def test_iam_live_token_server():
```
rmkeezer, line 297: Integration test added which checks to make sure the token is being retrieved

ibm_cloud_sdk_core/iam_token_manager.py
```diff
@@ -61,6 +62,7 @@ class IAMTokenManager(JWTTokenManager):
     """
     DEFAULT_IAM_URL = 'https://iam.cloud.ibm.com/identity/token'
```
padamstx, line 63: I think this should not include the path part, right?
```suggestion
    DEFAULT_IAM_URL = 'https://iam.cloud.ibm.com'
```

ibm_cloud_sdk_core/iam_token_manager.py
```diff
@@ -77,10 +79,15 @@ def __init__(self,
                  proxies: Optional[Dict[str, str]] = None,
                  scope: Optional[str] = None) -> None:
         self.apikey = apikey
-        self.url = url if url else self.DEFAULT_IAM_URL
+        if not url:
+            url = self.DEFAULT_IAM_URL
+        if url.endswith(self.OPERATION_PATH):
+            url = url[:-len(self.OPERATION_PATH)]
+        self.url = url
```
padamstx, line 86: note: at this point, `self.url` should be a valid IAM base URL string without the path part.

padamstx (approved):
Looks good. I suggested one small change to the default base url to remove the path part.  Will approve now to save another review cycle.

PR #109, opened 2021-03-10
Add get_query_param utility method to support pagination

mkistler:
This PR adds a small utility method to extract the value of a query parameter from a URL string. This will be used in the pagination support for Python.

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -17,11 +17,13 @@
 
 import os
 import datetime
-
 from typing import Optional
+import pytest
+
 from ibm_cloud_sdk_core import string_to_datetime, datetime_to_string, get_authenticator_from_environment
 from ibm_cloud_sdk_core import string_to_date, date_to_string
 from ibm_cloud_sdk_core import convert_model, convert_list
+from ibm_cloud_sdk_core import get_query_param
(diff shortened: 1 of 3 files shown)


padamstx (approved):
LGTM

test/test_utils.py
```diff
@@ -122,6 +125,51 @@ def test_date_conversion():
...
+    with pytest.raises(ValueError):
+        next = get_query_param(next_url, 'start')
+    # Duplicate param
+    next_url = '/api/v1/offerings?start=foo&start=bar&limit=10'
+    next = get_query_param(next_url, 'start')
+    assert next == 'foo'
+
+@pytest.mark.skipif(sys.version_info >= (3, 8), reason="Before Python 3.8, \
+     characters that affect netloc parsing under NFKC normalization were ignored.")
+def test_get_query_param_bad_url_legacy():
```
padamstx, line 162: I'm having a bit of trouble running this in my local environment.   For python 3.6 (3.6.13 to be precise), this test fails because the ValueError is being thrown similar to what is expected in the python 3.8+ case below.
Would it be better to just catch and absorb any ValueError that might be thrown inside the get_query_param method and just return None?

PR #110, opened 2021-03-17
feat: add datetime to string utils for lists

rmkeezer:
Adds two util functions to convert a list of DateTime to a list of str and vice versa. This is used for a new fix in the Python generator for formatting mock data values for generated unit tests.

### DIFF
--- a/ibm_cloud_sdk_core/__init__.py
+++ b/ibm_cloud_sdk_core/__init__.py
@@ -25,6 +25,8 @@ classes:
 functions:
     datetime_to_string: Serializes a datetime to a string.
     string_to_datetime: De-serializes a string to a datetime.
+    datetime_to_string_list: Serializes a list of datetimes to a list of strings.
+    string_to_datetime_list: De-serializes a list of strings to a list of datetimes.
     date_to_string: Serializes a date to a string.
     string_to_date: De-serializes a string to a date.
     convert_model: Convert a model object into an equivalent dict.
@@ -41,6 +43,7 @@ from .jwt_token_manager import JWTTokenManager
 from .cp4d_token_manager import CP4DTokenManager
 from .api_exception import ApiException
 from .utils import datetime_to_string, string_to_datetime, read_external_sources
+from .utils import datetime_to_string_list, string_to_datetime_list
 from .utils import date_to_string, string_to_date
 from .utils import convert_model, convert_list
 from .utils import get_query_param
--- a/ibm_cloud_sdk_core/utils.py
+++ b/ibm_cloud_sdk_core/utils.py
@@ -117,6 +117,40 @@ def string_to_datetime(string: str) -> datetime.datetime:
     return val.replace(tzinfo=datetime.timezone.utc)
 
 
+def string_to_datetime_list(string_list: List[str]) -> List[datetime.datetime]:
+    """De-serializes each string in a list to a datetime.
+
+    Args:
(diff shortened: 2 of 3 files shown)


ibm_cloud_sdk_core/utils.py
```diff
@@ -117,6 +117,38 @@ def string_to_datetime(string: str) -> datetime.datetime:
     return val.replace(tzinfo=datetime.timezone.utc)
 
 
+def string_list_to_datetime_list(string_list: List[str]) -> List[datetime.datetime]:
```
padamstx, line 120: Could we shorten the name to be `string_to_datetime_list`?

ibm_cloud_sdk_core/utils.py
```diff
@@ -117,6 +117,38 @@ def string_to_datetime(string: str) -> datetime.datetime:
...
+    """De-serializes each string in a list to a datetime.
+
+    Args:
+        string_list: list of strings containing datetime in iso8601 format.
+
+    Returns:
+        the de-serialized list of strings as a list of datetime objects.
+    """
+    if not isinstance(string_list, list):
+        return string_list
```
padamstx, line 130: I don't think we want to return the string list in case of an error do we?   The caller is expecting to get back a list of datetimes.  Could we throw an appropriate exception instead?

ibm_cloud_sdk_core/utils.py
```diff
@@ -117,6 +117,38 @@ def string_to_datetime(string: str) -> datetime.datetime:
...
+    datetime_list = []
+    for string_val in string_list:
+        datetime_list.append(string_to_datetime(string_val))
+    return datetime_list
+
+def datetime_list_to_string_list(datetime_list: List[datetime.datetime]) -> List[str]:
```
padamstx, line 136: Could we shorten the name to be `datetime_to_string_list`?

ibm_cloud_sdk_core/__init__.py
```diff
@@ -25,6 +25,8 @@
 functions:
     datetime_to_string: Serializes a datetime to a string.
     string_to_datetime: De-serializes a string to a datetime.
+    datetime_list_to_string_list: Serializes a list of datetimes to a list of strings.
```
padamstx, line 28: I think we also need `date_to_string_list` and `string_to_date_list`, don't we?

padamstx (changes requested):
I suggested shorter names for the functions, and I think we need functions for date as well.

ibm_cloud_sdk_core/utils.py
rmkeezer: Sure, updated with the shortened names

ibm_cloud_sdk_core/utils.py
rmkeezer: Updated with a ValueError being raised for invalid argument types

ibm_cloud_sdk_core/__init__.py
rmkeezer: Not needed for the python unit tests since the example value can be used without reformatting. The main reason datetime has an issue is because the python conversion of datetime to string uses `datetime.isoformat()` and ends up with some formatting differences to the original example. Whereas dates are simple converted to string `str(date)` so the unit tests don't need an extra util function for those

padamstx (approved):
LGTM

PR #111, opened 2021-04-01
fix: remove http retry support

christiancompton:
Recent codescans have flagged these two lines as vulnerabilities, given that `https://` should be used.  This PR removes these vulnerabilities, so that only retries with `https://` will be supported.

Ultimately, all of our users should be using `https://` in their service URLs.  Any objections to removing `http://` retry support?  CC: @padamstx @hudlow @rmkeezer


padamstx:
@christiancompton I'd like to understand more of the context around this.   I can sort of understand the code scanner not wanting us to use http://...  but we are merely setting things up IN CASE someone is trying to use an http://-style URL, and we have no control over whether users do that or not.   This seems like a false positive to me.  Can we just "allow-list" it somehow?

christiancompton:
Closing - I will address by allowlisting the false positives.

PR #115, opened 2021-05-14
fix(build): prevent semantic-release from committing package-lock.json

padamstx:
This PR removes package-lock.json and prevents semantic-release from committing it again in the future.


pyrooka (approved):
LGTM!

PR #116, opened 2021-06-03
test: add tests that explicitly set the Host request header

padamstx:
This PR just adds a few tests to verify that we can explicitly set the Host header in various ways such that it shows up in the resulting http request.


rmkeezer (approved):
LGTM!

PR #117, opened 2021-07-01
fix: remove reserved keywords from kwargs before passing it to requests

pyrooka:
This PR introduces a minor change/improvement: removal of some reserved keys from the `kwargs` of the base services' send function.

### DIFF
--- a/ibm_cloud_sdk_core/base_service.py
+++ b/ibm_cloud_sdk_core/base_service.py
@@ -277,6 +277,11 @@ class BaseService:
         # Check to see if the caller specified the 'stream' argument.
         stream_response = kwargs.get('stream') or False
 
+        # Remove the keys we set manually, don't let the user to overwrite these.
+        reserved_keys = ['method', 'url', 'headers', 'params', 'cookies']
+        for key in reserved_keys:
+            kwargs.pop(key, None)
+
         try:
             response = self.http_client.request(**request,
                                                 cookies=self.jar,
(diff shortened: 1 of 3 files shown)


rmkeezer (approved):
This approach seems fine for now. We will need to make sure the `reserved_keys` is correct in the future for any changes made to prepared_request.

The pylint is failing right now for what it seems to be an unrelated "no-member" error that I believe is a false positive. I assume this was added in a new version of pylint. I would go ahead and disable "no-member" in the pylintrc file and include it in this PR.

Also, I think it would be beneficial to add a unit test to sanity check this code, but I'll go ahead and approve this now.

ibm_cloud_sdk_core/base_service.py
```diff
@@ -277,6 +277,11 @@ def send(self, request: requests.Request, **kwargs) -> DetailedResponse:
         # Check to see if the caller specified the 'stream' argument.
         stream_response = kwargs.get('stream') or False
 
+        # Remove the keys we set manually, don't let the user to overwrite these.
+        reserved_keys = ['method', 'url', 'headers', 'params', 'cookies']
+        for key in reserved_keys:
+            kwargs.pop(key, None)
```
rmkeezer, line 283: I wonder if we want to just throw an error here or a message so the user knows their argument is being omitted.

pyrooka: I think throwing an exception is not a good idea, since removing the headers will be quite common. I've added a warning log for now.

padamstx (approved):
LGTM
After merging this PR, could you please do some additional testing with the platform-services python sdk (integration tests for a few services where you have the .env file, etc.)?

pyrooka:
Sure!

PR #118, opened 2021-08-05
refactor: use shared code for IAM based authenticators

pyrooka:
This PR introduces two new classes: `IAMRequestBasedTokenManager` and `IAMRequestBasedAuthenticator`. These contain common functionalities for IAM auth based flows.

I haven't changed the tests on purpose. (Well, there is a minor change to fix the codecov patch check). The reason is to make sure that no interface change has been made, so everything is working just like before.

### DIFF
--- a/ibm_cloud_sdk_core/authenticators/iam_request_based_authenticator.py
+++ b/ibm_cloud_sdk_core/authenticators/iam_request_based_authenticator.py
@@ -31,7 +31,6 @@ class IAMRequestBasedAuthenticator(Authenticator):
     Attributes:
         token_manager (TokenManager): Retrives and manages IAM tokens from the endpoint specified by the url.
     """
-    authentication_type = 'iam'
 
     def validate(self) -> None:
         """Validates the client_id, and client_secret for IAM token requests.
--- a/ibm_cloud_sdk_core/authenticators/iam_request_based_authenticator.py
+++ b/ibm_cloud_sdk_core/authenticators/iam_request_based_authenticator.py
@@ -20,6 +20,7 @@ from requests import Request
 
 from .authenticator import Authenticator
 
+
 class IAMRequestBasedAuthenticator(Authenticator):
     """The IAMRequestBasedAuthenticator class contains code that is common to all authenticators
     that need to interact with the IAM tokens service to obtain an access token.
--- a/ibm_cloud_sdk_core/authenticators/iam_request_based_authenticator.py
+++ b/ibm_cloud_sdk_core/authenticators/iam_request_based_authenticator.py
@@ -1,0 +1,113 @@
+# coding: utf-8
+
+# Copyright 2019 IBM All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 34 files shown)


pyrooka:
I am thinking about one more thing that could improve the project structure: put all the token managers into a single directory, just like the authenticators. Not sure would this break anything or not, I only found [this](https://github.com/IBM/platform-services-python-sdk/blob/main/ibm_platform_services/__init__.py#L19).
Screenshot 2021-08-05 at 19 13 29

padamstx:
> Not sure would this break anything or not,

if you move the token manager files to a new "token_managers" directory, couldn't you just update the ibm_cloud_sdk_core/__init__.py file to import them from the new location and effectively preserve the fact that they can be imported by users from the ibm_cloud_sdk_core package, like this?
```
from token_managers.iam_token_manager import IAMTokenManager
from token_managers.jwt_token_manager import JWTTokenManager
from token_managers.cp4d_token_manager import CP4DTokenManager
```

ibm_cloud_sdk_core/authenticators/iam_request_based_authenticator.py
```diff
@@ -0,0 +1,113 @@
...
+    that need to interact with the IAM tokens service to obtain an access token.
+
+    The bearer token will be sent as an Authorization header in the form:
+
+        Authorization: Bearer <bearer-token>
+
+    Attributes:
+        token_manager (TokenManager): Retrives and manages IAM tokens from the endpoint specified by the url.
+    """
+    authentication_type = 'iam'
```
padamstx, line 34: This won't apply to all IAM-based authenticators, so it probably belongs only in the IamAuthenticator class.

padamstx (approved):
Looks good!  I spotted one minor thing that needs to be changed but I'll approve now to avoid a re-review.   It looks like this refactoring work will greatly reduce the amount of new code needed for the ContainerAuthenticator and (later on) the VPCInstanceAuthenticator.   I've found the same thing in the Java core as well and I suspect @dpopp07 is seeing the same in the Node core.  @dpopp07 Thank you for the idea to do this refactoring :)

dpopp07 (approved):
I second Phil's comment but otherwise, looks good! 👍

pyrooka:
I've pushed 3 more commits since your review.
The first 2 contain the `authentication_type = 'iam'` fix and the restructuring of the token managers.
The last commit is really just about the code style. I found that there was a bit inconsistency so I checked every single file, reordered/grouped the imports and added missing blank lines (where it was necessary).

More info:
- Surround top-level function and class definitions with two blank lines: https://www.python.org/dev/peps/pep-0008/#blank-lines
- Imports grouping and ordering: https://www.python.org/dev/peps/pep-0008/#imports

ibm_cloud_sdk_core/token_managers/iam_request_based_token_manager.py
```diff
@@ -95,7 +96,7 @@ def request_token(self) -> dict:
              A dictionary containing the bearer token to be subsequently used service requests.
         """
         headers = {
-            'Content-type': 'applicaton/x-www-form-urlencoded',
+            'Content-type': 'application/x-www-form-urlencoded',
```
padamstx, line 99: wow, the misspelled "applicaton" didn't cause a problem before???  :)

padamstx (approved):
Latest changes LGTM!

ibm_cloud_sdk_core/token_managers/iam_request_based_token_manager.py
pyrooka: I was surprised too! The container authenticator tests will cover this, so I didn't change the tests here.

PR #121, opened 2021-08-17
fix: encode serialized JSON string as UTF-8

pyrooka:
This PR adds encoding to the serialized JSON payload. We are already doing this if the payload is passed as a string.

In the generated SDKs, for example in the [Platform Services/Case Management](https://github.com/IBM/platform-services-python-sdk/blob/main/ibm_platform_services/case_management_v1.py#L217), we serialize the `data` dictionary and pass it to the `prepare_request` function as a string, therefore I think it's safe to add encoding here.

### DIFF
--- a/ibm_cloud_sdk_core/base_service.py
+++ b/ibm_cloud_sdk_core/base_service.py
@@ -18,7 +18,6 @@ import gzip
 import json as json_import
 import logging
 import platform
-import sys
 from http.cookiejar import CookieJar
 from os.path import basename
 from typing import Dict, List, Optional, Tuple, Union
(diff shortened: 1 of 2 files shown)


ibm_cloud_sdk_core/base_service.py
```diff
@@ -373,14 +372,13 @@ def prepare_request(self,
         params = cleanup_values(params)
         request['params'] = params
 
-        if sys.version_info >= (3, 0) and isinstance(data, str):
```
pyrooka, line 376: We are not supporting Python 2 as far as I know, so I removed the version check.

ibm_cloud_sdk_core/base_service.py
```diff
@@ -373,14 +372,13 @@ def prepare_request(self,
...
+        if isinstance(data, str):
```
pyrooka, line 375: We don't support Python 2 as far as I know, so I removed the version check.

padamstx (approved):
LGTM

dpopp07 (approved):
Looks good! 👍

PR #122, opened 2021-08-20
fix(retries): change default retry_interval to 1 second (was 0.1)

padamstx:
This commit changes the default value of the retry_interval
parameter (in the BaseService.enable_retries method) from 0.1
to 1.0.   This will result in default retry intervals
of 1, 2, 4, 8, ... seconds instead of 0.1, 0.2, 0.4, 0.8, ... seconds.
This change will align the python core retry support with Go.
Also added some detailed info in the docstring for the method as well.


rmkeezer (approved):
LGTM

PR #130, opened 2021-11-11
fix: strip trailing slashes in BaseService.set_service_url

padamstx:
This commit modifies the BaseService.set_service_url()
function so that it removes any trailing slashes.

Previously, a service URL containing a trailing slash
(e.g. "https://myserver.com/api/v1/") would result in a
request URL that contains two slashes between the service URL
and the operation path (e.g. "https://myserver.com/api/v1//myoperationpath"),
and some server implementations might fail to process such
a request properly.

With the changes in this commit, a request URL of
"https://myserver.com/api/v1/myoperationpath" will be used instead.

### DIFF
--- a/test/test_base_service.py
+++ b/test/test_base_service.py
@@ -20,7 +20,6 @@ from ibm_cloud_sdk_core import CP4DTokenManager
 from ibm_cloud_sdk_core import get_authenticator_from_environment
 from ibm_cloud_sdk_core.authenticators import (IAMAuthenticator, NoAuthAuthenticator, Authenticator,
                                                BasicAuthenticator, CloudPakForDataAuthenticator)
-from ibm_cloud_sdk_core.utils import strip_extra_slashes
 
 
 class IncludeExternalConfigService(BaseService):
@@ -698,6 +697,7 @@ def test_user_agent_header():
     assert response.get_result().request.headers.__getitem__(
         'user-agent') == user_agent_header['User-Agent']
 
+
 @responses.activate
 def test_reserved_keys(caplog):
     service = AnyServiceV1('2021-07-02', authenticator=NoAuthAuthenticator())
@@ -722,6 +722,7 @@ def test_reserved_keys(caplog):
     assert caplog.record_tuples[1][2] == '"url" has been removed from the request'
     assert caplog.record_tuples[2][2] == '"headers" has been removed from the request'
     assert caplog.record_tuples[3][2] == '"cookies" has been removed from the request'
+
 
 @responses.activate
 def test_ssl_error():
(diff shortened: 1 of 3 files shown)


test/test_base_service.py
```diff
@@ -810,56 +811,71 @@ def test_json():
     service = AnyServiceV1('2018-11-20', authenticator=NoAuthAuthenticator())
     req = service.prepare_request('POST', url='', headers={
                                   'X-opt-out': True}, data={'hello': 'world', 'fóó': 'bår'})
-    assert req.get('data') == b'{"hello": "world", "f\\u00f3\\u00f3": "b\\u00e5r"}'
+    assert req.get(
+        'data') == b'{"hello": "world", "f\\u00f3\\u00f3": "b\\u00e5r"}'
 
 
-def test_trailing_slash():
```
padamstx, line 816: moved this to test_utils.py

test/test_base_service.py
```diff
@@ -810,56 +811,71 @@ def test_json():
...
-    assert strip_extra_slashes('/////') == '/'
-    assert strip_extra_slashes('https://host') == 'https://host'
-    assert strip_extra_slashes('https://host/') == 'https://host/'
-    assert strip_extra_slashes('https://host//') == 'https://host/'
-    assert strip_extra_slashes('https://host/path') == 'https://host/path'
-    assert strip_extra_slashes('https://host/path/') == 'https://host/path/'
-    assert strip_extra_slashes('https://host/path//') == 'https://host/path/'
-    assert strip_extra_slashes('https://host//path//') == 'https://host//path/'
-
+def test_service_url_handling():
```
padamstx, line 818: Removed the direct calls to strip_extra_slashes() so this test will focus on the service and service_url

test/test_base_service.py
```diff
@@ -810,56 +811,71 @@ def test_json():
...
+
     service.set_service_url('https://host/')
-    assert service.service_url == 'https://host/'
+    assert service.service_url == 'https://host'
+
     req = service.prepare_request('POST',
                                   url='/path/',
                                   headers={'X-opt-out': True},
                                   data={'hello': 'world'})
-    assert req.get('url') == 'https://host//path/'
```
padamstx, line 837: we can no longer have double-slashes between the service_url and the operation path

padamstx:
Reminder to myself:  before we merge this, we need to test this version of the core with existing generated unit tests.

padamstx:
I also did a test with the platform-services python SDK where I ran the global-search integration test with and without the python core fix.
The service URL I used in each case was: `https://api.global-search-tagging.cloud.ibm.com/`.
In the "before" scenario (without the python core fix), I see this in the http debug messages:
```
send: b'POST //v3/resources/search?limit=1 HTTP/1.1\r\nHost: api.global-search-tagging.cloud.ibm.com
```
so the path value passed in the HTTP request message does in fact have a leading double-slash (the global-search server still processes this request just fine, but not all servers behave this way).

In the "after" scenario (with the python core fix), I see this in the http debug messages:
```
send: b'POST /v3/resources/search?limit=1 HTTP/1.1\r\nHost: api.global-search-tagging.cloud.ibm.com
```
so the path value now has just a single slash.

rmkeezer (approved):
LGTM

PR #131, opened 2021-11-15
build: use bionic dist in travis

padamstx:
This PR fixes a recent pylint problem seen with python 3.7.1 (the default version used by the travis xenial build image) by upgrading the build up to the bionic image where python 3.7.6 is used instead.
I also took this opportunity to remove the use of tox in the travis build since it's really not needed.  Instead, we just run the "make ci" command in each of the travis build jobs (one job per python version... 3.6 - 3.9).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+dist: bionic
+
 language: python
 
 matrix:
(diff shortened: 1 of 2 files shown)


rmkeezer (approved):
LGTM!

PR #132, opened 2021-11-15
build: bump requests and urllib3 deps

padamstx:
This PR bumps up our requests dependency to 2.26.0, and explicitly adds a dependency on urllib3 (>= 1.26.0) due to the retry-related code in the python core.


rmkeezer (approved):
LGTM

PR #133, opened 2021-11-15
build: remove tox.ini

padamstx:
This PR simply removes tox.ini since we no longer use Tox in our travis builds.


rmkeezer (approved):
LGTM

PR #135, opened 2021-12-16
fix: pylint simplifications

rmkeezer:
Pylint updated and is now failing. These simplifications fix the lint errors.


padamstx (approved):
LGTM

PR #137, opened 2022-02-23
build: upgrade pytest

pyrooka:
This PR updates pytest to the next major version (7.x.x) and pins the `setuptools` package version in the Travis deploy step to fix publishing to PyPi.

~~NOTE: Python 3.6 is no longer supported officially, but it seems to work.~~


padamstx (approved):
LGTM
I noticed that the build status was not current in the PR "Conversation" tab, so I restarted both the PR and Branch builds in Travis and they were both clean, even though the status doesn't show up here

PR #138, opened 2022-02-24
build: update Python versions

pyrooka:
This PR drops the support for Python 3.6 and adds support for Python 3.10.


padamstx (approved):
LGTM

padamstx:
@pyrooka I think we might want to use a "feat" type commit for this change to sort of draw attention to the change in supported python versions.  What do you think?

PR #139, opened 2022-03-08
chore: set minimum TLS version to v1.2

pyrooka:
This PR introduces a custom HTTP adapter that wraps the original one
and adds SSL context, that sets the minimum TLS version required to v1.2.

### DIFF
--- a/ibm_cloud_sdk_core/utils.py
+++ b/ibm_cloud_sdk_core/utils.py
@@ -23,6 +23,7 @@ from typing import List, Union
 from urllib.parse import urlparse, parse_qs
 
 from requests.adapters import HTTPAdapter
+from urllib3.util.ssl_ import create_urllib3_context
 
 import dateutil.parser as date_parser
 
@@ -37,7 +38,9 @@ class SSLHTTPAdapter(HTTPAdapter):
     def init_poolmanager(self, connections, maxsize, block):
         """Extends the parent's method by adding minimum SSL version to the args.
         """
-        super().init_poolmanager(connections, maxsize, block, ssl_minimum_version=ssl.TLSVersion.TLSv1_2)
+        ssl_context = create_urllib3_context()
+        ssl_context.minimum_version = ssl.TLSVersion.TLSv1_2
+        super().init_poolmanager(connections, maxsize, block, ssl_context=ssl_context)
 
 
 def has_bad_first_or_last_char(val: str) -> bool:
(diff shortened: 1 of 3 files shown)


padamstx (approved):
LGTM.
We just need to make really sure that this will also work with retries enabled.

ibm_cloud_sdk_core/utils.py
```diff
@@ -16,14 +16,30 @@
...
+    """Wraps the original HTTP adapter and adds additional SSL context.
+    """
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    #pylint: disable=arguments-differ
+    def init_poolmanager(self, connections, maxsize, block):
+        """Extends the parent's method by adding minimum SSL version to the args.
+        """
+        super().init_poolmanager(connections, maxsize, block, ssl_minimum_version=ssl.TLSVersion.TLSv1_2)
```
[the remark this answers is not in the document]
pyrooka: Yeah that's right, we need to use `ssl_context`! I've already made this change, but haven't pushed it yet, because I want to do some further testing to make sure everything works fine.
By the way, I found it here: https://github.com/urllib3/urllib3/blob/main/src/urllib3/poolmanager.py#L53, but it doesn't work, because this is the main branch (future 2.0 release) and this option is missing from the current release branch (1.26.x).

ibm_cloud_sdk_core/utils.py
```diff
@@ -16,14 +16,33 @@
...
+    """Wraps the original HTTP adapter and adds additional SSL context.
+    """
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    #pylint: disable=arguments-differ
+    def init_poolmanager(self, connections, maxsize, block):
+        """Extends the parent's method by adding minimum SSL version to the args.
+        """
+        ssl_context = create_urllib3_context()
```
ricellis, line 41: I was trying to understand the difference between this and `ssl.create_default_context()`. It appears the `urllib3` version is modified to remove some insecure older things so it seems like using this route to get a context is good anyway. Was there any other reason?

ricellis (approved):
I validated that mounting the SSLHTTPAdapter for `http` as well as `https` doesn't seem to cause any problems. LGTM I have just one question arond the SSL context, but that's more for my understanding than any impediment to merging this.

ibm_cloud_sdk_core/utils.py
pyrooka: No other reason. I found it a bit more secure, that's all.

PR #142, opened 2022-06-06
fix: avoid warning when custom headers are used

padamstx:
Fixes: arf/planning-sdk-squad#3250

This PR avoids a warning being logged when the 'headers' key is removed from kwargs by the BaseService.send() method.


dpopp07 (approved):
Looks good! 👍

PR #146, opened 2022-06-28
build: used fixed version of setuptools for Python 3.7 and 3.8

pyrooka:
This commit fixes the version of the `setuptools` package for Python 3.7
and 3.8 builds to address build issues.


padamstx (approved):
LGTM

PR #147, opened 2022-07-27
feat(CP4D Authentication): add ssl verification for self-signed certificates

LBExposito:
These modifications allow the user to have a self-signed certificate in the SSL verification process when authenticating to an STT instance on CP4D.

The `verify` attribute is used by the [requests python module](https://pypi.org/project/requests/) to indicate the path to a custom CA Bundle ([resquests documentation](https://requests.readthedocs.io/en/latest/api/?highlight=verify)).

With these changes, it wont be necessary to disable ssl verification anymore when using a self-signed certificate with STT on CP4D.

### DIFF
--- a/ibm_cloud_sdk_core/authenticators/cp4d_authenticator.py
+++ b/ibm_cloud_sdk_core/authenticators/cp4d_authenticator.py
@@ -59,14 +59,15 @@ class CloudPakForDataAuthenticator(Authenticator):
                  apikey: str = None,
                  disable_ssl_verification: bool = False,
                  headers: Optional[Dict[str, str]] = None,
-                 proxies: Optional[Dict[str, str]] = None) -> None:
+                 proxies: Optional[Dict[str, str]] = None,
+                 verify: Optional[str] = None) -> None:
         # Check the type of `disable_ssl_verification`. Must be a bool.
         if not isinstance(disable_ssl_verification, bool):
             raise TypeError('disable_ssl_verification must be a bool')
 
         self.token_manager = CP4DTokenManager(
             username=username, password=password, apikey=apikey, url=url,
-            disable_ssl_verification=disable_ssl_verification, headers=headers, proxies=proxies)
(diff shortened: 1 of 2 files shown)


bobfang:
to elaborate a bit more on this, these changes are for the Authentication step (getting a token). During regular SDK calls we can still pass `verify` as a kwarg and get it passed through to Requests.

ibm_cloud_sdk_core/authenticators/cp4d_authenticator.py
```diff
@@ -59,14 +59,15 @@ def __init__(self,
                  apikey: str = None,
                  disable_ssl_verification: bool = False,
                  headers: Optional[Dict[str, str]] = None,
-                 proxies: Optional[Dict[str, str]] = None) -> None:
+                 proxies: Optional[Dict[str, str]] = None,
+                 verify: Optional[str] = None) -> None:
```
padamstx, line 63: Please also document the `verify` parameter in the docstrings above, similar to headers, proxies, etc.

ibm_cloud_sdk_core/token_managers/cp4d_token_manager.py
```diff
@@ -57,9 +57,11 @@ def __init__(self,
                  apikey: str = None,
                  disable_ssl_verification: bool = False,
                  headers: Optional[Dict[str, str]] = None,
-                 proxies: Optional[Dict[str, str]] = None) -> None:
+                 proxies: Optional[Dict[str, str]] = None,
+                 verify: Optional[str] = None) -> None:
```
padamstx, line 61: Please document `verify` in the docstrings above.

padamstx (changes requested):
Change look good, but we just need to add "verify" to the docstrings of each function that it was added to...

LBExposito:
Docstrings are updated !

padamstx (approved):
LGTM

PR #161, opened 2023-04-03
feat: return response text as result intead of the whole object

pyrooka:
In case of text responses like `text/plain`, `text/html`, etc... the `send`
method returns the whole response object as the result in the 
`DetailedResponse` object. This PR changes this behavior so that
the `send` method will return the response text from the body as string.
It means it can be accessed directly from the response by calling
`response.get_result()`.

### DIFF
--- a/ibm_cloud_sdk_core/base_service.py
+++ b/ibm_cloud_sdk_core/base_service.py
@@ -40,6 +40,7 @@ from .utils import (
     read_external_sources,
     strip_extra_slashes,
     SSLHTTPAdapter,
+    TextResult,
 )
 from .version import __version__
 
@@ -330,7 +331,7 @@ class BaseService:
         elif stream_response:
             result = response
         elif is_text_mimetype(content_type):
-            result = response.text
+            result = TextResult(response)
         elif is_json_mimetype(content_type):
             # If this is a JSON response, then try to unmarshal it.
             try:
--- a/ibm_cloud_sdk_core/base_service.py
+++ b/ibm_cloud_sdk_core/base_service.py
@@ -34,6 +34,7 @@ from .token_managers.token_manager import TokenManager
 from .utils import (
     has_bad_first_or_last_char,
     is_json_mimetype,
+    is_text_mimetype,
     remove_null_values,
     cleanup_values,
     read_external_sources,
@@ -311,6 +312,7 @@ class BaseService:
                     logger.warning('"%s" has been removed from the request', key)
         try:
             response = self.http_client.request(**request, cookies=self.jar, **kwargs)
+            content_type = response.headers.get('Content-Type')
 
             # Process a "success" response.
             if 200 <= response.status_code <= 299:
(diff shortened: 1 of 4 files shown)


padamstx:
The changes in this PR will require an SDK user to change their code in terms of how a string-based response is received.   The app will no longer need to explicitly access the `text` field of the response, since that's what's now being set in the "result" field.   This constitutes a breaking change from the perspective of an app using a Python SDK.
We'll need to tread very carefully here and consider the ramifications of this change.

pyrooka:
Agree, we should be careful. I treat this PR as a POC for now due to the breaking change it introduces.

ibm_cloud_sdk_core/base_service.py
```diff
@@ -309,40 +311,43 @@ def send(self, request: requests.Request, **kwargs) -> DetailedResponse:
...
+        content_type = response.headers.get('Content-Type')
+
+        # Process a "success" response.
+        if response.status_code == 204 or request['method'] == 'HEAD' or not response.text:
+            # There is no body content for a HEAD response, a 204 response or no content at all.
+            result = None
+        elif stream_response:
+            result = response
+        elif is_text_mimetype(content_type):
+            result = TextResult(response)
```
pyrooka, line 334: The new "wrapper" class to add more backward compatibility.

ibm_cloud_sdk_core/utils.py
```diff
@@ -409,6 +410,36 @@ def is_json_mimetype(mimetype: str) -> bool:
...
+    Keyword Args:
+        response: The response to the service request.
+    """
+
+    def __init__(self, response: Response) -> None:
+        # Init parent.
+        super().__init__()
+        self.__dict__.update(response.__dict__)
+        # Overrides parent's `__repr__` method.
+        self.__repr__ = self.__repr__
```
pyrooka, line 442: We need to copy all the parent attributes then overwrite the `__repr__` method to return our custom value.

test/test_base_service.py
```diff
@@ -432,8 +432,8 @@ def test_request_success_nonjson():
     service = AnyServiceV1('2018-11-20', authenticator=NoAuthAuthenticator())
     prepped = service.prepare_request('GET', url='')
     detailed_response = service.send(prepped)
-    # It's odd that we have to call ".text" to get the string value
-    # (see issue 3557)
+    assert repr(detailed_response.get_result()) == '<h1>Hola, amigo!</h1>'
+    assert str(detailed_response.get_result()) == '<h1>Hola, amigo!</h1>'
```
pyrooka, line 436: Still not perfect... `repr` or `str` has to be called to get the string value. It's sad, because in the test output it almost looks good
```
AssertionError: assert <h1>Hola, amigo!</h1> == '<h1>Hola, amigo!</h1>'
```
but the first one is actually an instance of the `TextResult` object and that's just the representation. That means we can't
```
'Result:' + response.get_result()
```
but
`'Result:' + str(response.get_result())` or `'Result: {0}'.format(response.get_result())`

pyrooka (commented):
A few comments on my last commit.

pyrooka:
We agreed on not to merge this feature due to the breaking changes and keep the current behavior, so I'm closing this PR now but won't delete the feature branch in case of we need it in the future.

PR #165, opened 2023-05-04
fix(ContainerAuthenticator): add sa-token as default CR token filename

padamstx:
This commit adds support for a second default CR token filename. If the user doesn't specify a CR token filename, the authenticator will first try '/var/run/secrets/tokens/vault-token', and then '/var/run/secrets/tokens/sa-token' in that order.

### DIFF
--- a/ibm_cloud_sdk_core/token_managers/container_token_manager.py
+++ b/ibm_cloud_sdk_core/token_managers/container_token_manager.py
@@ -80,7 +80,8 @@ class ContainerTokenManager(IAMRequestBasedTokenManager):
         This can be used to obtain an access token with a specific scope.
     """
 
-    DEFAULT_CR_TOKEN_FILENAME = '/var/run/secrets/tokens/vault-token'
+    DEFAULT_CR_TOKEN_FILENAME1 = '/var/run/secrets/tokens/vault-token'
+    DEFAULT_CR_TOKEN_FILENAME2 = '/var/run/secrets/tokens/sa-token'
 
     def __init__(
         self,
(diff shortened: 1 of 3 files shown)


ibm_cloud_sdk_core/token_managers/container_token_manager.py
```diff
@@ -177,3 +178,15 @@ def set_iam_profile_id(self, iam_profile_id: str) -> None:
             iam_profile_id: id of the linked trusted IAM profile to be used when obtaining the IAM access token
         """
         self.iam_profile_id = iam_profile_id
+
+    def read_file(self, filename: str) -> str:
+        """Read in the specified file and return the contents as a string"""
```
pyrooka, line 183: It seems the linter doesn't complain about missing docstrings, but I have a suggestion. It's up to you to change it or not. :)
```suggestion
        """Read in the specified file and return the contents as a string.

        Args:
            filename: ID of the linked trusted IAM profile to be used when obtaining the IAM access token.

        Returns:
            A string with the contents of the file.

        Raises:
            Exception: An error occured reading the file.
        """
```

pyrooka (approved):
Looks good!

ibm_cloud_sdk_core/token_managers/container_token_manager.py
padamstx: Good idea, but I'll use a different description for the filename arg :joy:
Sorry, couldn't resist :)

padamstx: fixed

pyrooka: LOL, you got me! I was too lazy and copied the docstring from another method. 😂

dpopp07 (approved):
Looks good and tested with internal cluster 👍

PR #175, opened 2023-10-02
feat: bump min version of python 3.7 -> 3.8

padamstx:
This commit modifies the project to bump our min
supported version of python from 3.7 to 3.8,
and add 3.11 to the versions that we build with.
This is being done since python 3.7 is no longer
in service (https://devguide.python.org/versions/).


setup.py
```diff
@@ -31,13 +31,15 @@
     os.system('python setup.py sdist upload -r pypi')
     sys.exit()
 
-with open('requirements.txt') as f:
+with open('requirements.txt', encoding='utf-8') as f:
```
padamstx, line 34: these are just fixes to linter warnings :)

requirements-dev.txt
```diff
@@ -1,7 +1,6 @@
-coverage>=4.5.4
-pylint>=2.6.0,<3.0.0
-pytest>=7.0.1,<8.0.0
-pytest-cov>=2.2.1,<3.0.0
-responses>=0.12.1,<1.0.0
-tox>=3.2.0,<4.0.0
-black>=22.10
+coverage>=7.3.2,<8.0.0
```
padamstx, line 1: went ahead and freshened up our dependencies as well :)

CONTRIBUTING.md
```diff
@@ -34,7 +34,7 @@ If you want to contribute to the repository, here's a quick guide:
     * Only use spaces for indentation.
     * Create minimal diffs - disable on save actions like reformat source code or organize imports. If you feel the source code should be reformatted create a separate PR for this change.
     * Check for unnecessary whitespace with `git diff --check` before committing.
-    * Make sure your code supports Python 3.7, 3.8, 3.9 and 3.10. You can use `pyenv` and `tox` for this
+    * Make sure your code tests clean on the project's supported versions of Python.  You can use the `venv` module to create virtual environments for this.
```
pyrooka, line 37: Good thinking! This way we don't have to maintain the Python versions in this doc. :)

pyrooka (approved):
I like all the small changes! :)

PR #177, opened 2023-10-09
fix: use on-the-fly compression only for file objects

pyrooka:
When we introduced the new `GzipStream` utility class which
handles the on-the-fly compression of the file like object.
We also started using it for in-memory objects, like bytes,
but it has introduced a minor incompatibility with the existing SDKs.
This commit contains the fix for that. From now the new utility class
will be used only for file-like objects and the rest of the types will
be handled as before (compressing in memory).


dpopp07 (approved):
Looks good! 👍

PR #179, opened 2023-11-06
update urllib to latest

hemanthrough:
bumping urllib due to https://github.com/advisories/GHSA-g4mx-q9vg-27p4

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 requests>=2.31.0,<3.0.0
-urllib3>=2.0.7
+urllib3>=1.26.18,<2.0.0
 python_dateutil>=2.8.2,<3.0.0
 PyJWT>=2.8.0,<3.0.0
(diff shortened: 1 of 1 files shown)


pyrooka:
Hi! Thanks for your contribution, but we would like to avoid upgrading to the next major version (v2) at the moment. 
As I can see, the CVE got fixed in `1.26.18` too, so maybe all we need to do is replacing the minimum version with that.

hemanthrough:
@pyrooka sure will do that 
is there any reason for not doing that ?

pyrooka:
Well, we plan to do it in the near future, but we have other projects that also need to be updated and we would like to do it in one batch. So the main reason is our internal tracking and to keep these projects in sync.

hemanthrough:
ok thanks.
I suppose you will take it from here :) ? (wrt to travis and stuff) 
Should I do anything else ?

padamstx:
Closing this PR and opened [this PR](https://github.com/IBM/python-sdk-core/pull/180) due to Travis issues.

PR #182, opened 2023-11-22
feat: improve behavior of HTTP redirects

pyrooka:
This commit modifies the Python core so that it will include "safe" headers
when performing a cross-site redirect where both the original and redirected
hosts are within IBM's "cloud.ibm.com" domain.


pyrooka:
I've converted this PR to to draft, to prevent merging it before the necessary CISO approvals.

ibm_cloud_sdk_core/base_service.py
```diff
@@ -314,6 +323,38 @@ def send(self, request: requests.Request, **kwargs) -> DetailedResponse:
...
+            # Check if the response is a redirect to another host.
+            while response.is_redirect and response.next is not None and redirects_count < MAX_REDIRECTS:
+                redirects_count += 1
+
+                # urllib3 has already prepared a request that can almost be used as-is.
+                next_request = response.next
+
+                # If both the original and the redirected URL are under the `.cloud.ibm.com` domain,
+                # copy the safe headers that are used for authentication purposes,
+                if self.service_url.endswith('.cloud.ibm.com') and urlparse(next_request.url).netloc.endswith(
```
padamstx, line 337: `self.service_url.endswith('.cloud.ibm.com')` will most definitely not work in the general case. To see why, consider a service url like `https://myhost.ibm.com/api` or `https://myhost.ibm.com:8080`.  I think you'll need to parse `self.service_url` and extract the hostname from it.

ibm_cloud_sdk_core/base_service.py
```diff
@@ -314,6 +323,38 @@ def send(self, request: requests.Request, **kwargs) -> DetailedResponse:
...
+                # otherwise remove them manually, because `urllib3` doesn't strip all of them.
+                else:
+                    for header in SAFE_HEADERS:
+                        next_request.headers.pop(header, None)
+
+                response = self.http_client.send(next_request, **kwargs)
+
+            # If we reached the max number of redirects and the last response is still a redirect
+            # stop processing the response and return an error to the user.
+            if redirects_count == MAX_REDIRECTS and response.is_redirect:
```
padamstx, line 353: Why not do the limit check immediately after incrementing `redirect_count` above?  I think it could be cleaner as you would just need to check:
`if redirects_count >= MAX_REDIRECTS:`

test/test_base_service.py
```diff
@@ -788,6 +788,166 @@ def test_retry_config_external():
     assert retry_err.value.reason == error
 
 
+@responses.activate
```
padamstx, line 791: Need to add some additional tests:
* redirect to same host (both inside cloud.ibm.com)
* redirect to same host (both outside cloud.ibm.com)

Also, because you are completely overriding the default "follow redirects" behavior, I think you'll need to add some tests for redirect scenarios involving POST and a representative sample of redirect status codes.   There are some subtle gotcha's when implementing redirects and we need to make sure that we are providing equivalent function as compared to the default "follow redirects" behavior.
I had to do this in the Java implementation as well.  For example, for 301/302 status codes, a POST should be re-written as a GET with no body (it may fail in the redirected server, but that's what the HTTP semantics should be).  On the other hand, 307/308 status codes should NOT cause a POST to be re-written as a GET.
You should study the code that you are overriding and try to match it as closely as possible other than adding support for the *.cloud.ibm.com stuff.

padamstx (commented):
I think this is a good start, but I think there are some missing pieces to the implementation.

ibm_cloud_sdk_core/base_service.py
pyrooka: Yeah, that's a good point and I already parse the url in the second part of the expression... :)

ibm_cloud_sdk_core/base_service.py
pyrooka: Makes sense! I need to excuse myself so: initially I implemented the error handling in a different way then I forgot to refactor that check.. 😄

test/test_base_service.py
pyrooka: Tests are already there those scenarios. Copied from the tests
```python
def test_redirect_ibm_to_ibm_success():
    url_from = 'http://region1.cloud.ibm.com/'
    url_to = 'http://region2.cloud.ibm.com/'
    ...
```
and
```python
def test_redirect_not_ibm_to_not_ibm_fail():
    url_from = 'http://region1.notcloud.ibm.com/'
    url_to = 'http://region2.notcloud.ibm.com/'
    ...
```
Are you talking about something else?

pyrooka: > Also, because you are completely overriding the default "follow redirects" behavior...

That's not the case. Still the `requests` package handles the redirections, the `"allow_redirect": False"` option only prevents to fire the next request automatically, but it's still being processed and resolved appropriately. From the `session.py` in the `requests` package:
```python
# If redirects aren't being followed, store the response on the Request for Response.next().
        if not allow_redirects:
            try:
                r._next = next(
                    self.resolve_redirects(r, request, yield_requests=True, **kwargs)
                )
            except StopIteration:
                pass
```
Let's touch base later, to make sure we are on the same page! :)

padamstx: > Are you talking about something else?

1. redirect to same host (both inside cloud.ibm.com):
from: region1.cloud.ibm.com
to:   region1.cloud.ibm.com

2. redirect to same host (both outside cloud.ibm.com):
from: region1.notcloud.ibm.com
to: region1.notcloud.ibm.com

Redirecting to the same host SHOULD allow the safe headers to be included in the redirected request, regardless of whether the host is inside or outside the cloud.ibm.com domain.

padamstx: > Let's touch base later, to make sure we are on the same page!

Yes, let's discuss tomorrow.

padamstx (approved):
Additional testcases look good!

PR #184, opened 2023-12-11
fix: use retry_interval as retry backoff_max

padamstx:
This commit makes a subtle change to the "retry_interval" argument of BaseService.enable_retries().
It is now used to supply the value for the "backoff_max" argument passed to the urllib3's Retry ctor, and now represents the maximum time to wait before the next retry.


pyrooka (approved):
Looks good to me!

dpopp07 (approved):
Looks good!

PR #185, opened 2024-01-18
fix: more consistent naming of HTTP status code related attributes

pyrooka:
This commit fixes a small inconsistency in the naming
of the HTTP status code attributes between the successful
and the error responses.

### DIFF
--- a/ibm_cloud_sdk_core/api_exception.py
+++ b/ibm_cloud_sdk_core/api_exception.py
@@ -14,6 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import warnings
 from http import HTTPStatus
 from typing import Optional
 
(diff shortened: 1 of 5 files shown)


ibm_cloud_sdk_core/api_exception.py
```diff
@@ -39,15 +40,28 @@ def __init__(self, code: int, *, message: Optional[str] = None, http_response: O
...
+    def code(self):
+        """The old `code` property with a deprecation warning."""
+
+        warnings.warn(
+            'Using the `code` attribute on the `ApiException` is deprecated and'
+            'will be removed in the future. Use `status_code` instead.',
+            DeprecationWarning,
+        )
+        return self.status_code
+
```
pyrooka, line 62: I thought it would be good to not remove the old property completely, but mark it deprecated and give some time to the the users to update their code. Also, the date in the comment is arbitrary, there is no real reason behind it. :)

padamstx: By renaming the field from `code` to `status_code` above, I think you've introduced a breaking change despite the addition of this new getter function.
What I mean is...  previously a user would need to access the `code` field like this:
```
    print('The status code is: ', exception.code)
```
That will no longer work as the user would need to do one of the following instead now:
```
    print('The status code is: ', exception.status_code)
```
or
```
    print('The status code is: ', exception.code())
```

IOW, I don't think the user can continue to just use `exception.code`, can they???

test/test_base_service.py
```diff
@@ -398,7 +398,7 @@ def test_request_success_invalid_json():
         service = AnyServiceV1('2018-11-20', authenticator=NoAuthAuthenticator())
         prepped = service.prepare_request('GET', url='')
         service.send(prepped)
-    assert err.value.code == 200
+    assert err.value.status_code == 200
```
padamstx, line 401: I think it would be good to have a testcase that still uses the old way `err.value.code` to make sure that still works (see my previous comment about this above)

padamstx (commented):
I left a couple of comments re: whether this is a breaking change or not.  Let's resolve this before merging.

test/test_base_service.py
pyrooka: Good idea, I'll add some!

ibm_cloud_sdk_core/api_exception.py
pyrooka: Fortunately that's not the case. In short, the `@property` decorator creates a valid attribute (well only with a setter by default), so it can be used as before. See the example below;
```
>>> ex = ApiException(404, message="Not found...")
>>> ex.code
404
>>> ex.status_code
404
```
The only difference is, when calling the `code` property, a deprecation message will be printed to the `stderr` at warning level.

padamstx: Aha! I didn't notice the `@property` decorator, and I probably wouldn't have known what that was for even if I did :joy:.

Issue #186, opened 2024-01-23
set_disable_ssl_verification not applying


nrocaalh:
Hi, 

I have configured the **_set_disable_ssl_verification_** to **_True_** but it's not applying, and this error is of certificate not found is persistent

{SSLError(MaxRetryError("HTTPSConnectionPool(host='iam.cloud.ibm.com', port=443): Max retries exceeded with url: /identity/token (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1123)')))"))}

I'm applying the disable ssl verification following this documentation https://github.com/IBM/ibm-cloud-sdk-common/blob/main/README.md#programmatically-1.

Thank you,

pyrooka:
Thanks for the feedback, I'll look into it. Could you attach the full trace back of the error? It would help a lot.

pyrooka:
@nrocaalh we just merged the linked PR which should fix the issue so I am closing this, but feel free to re-open it if needed.

nrocaalh:
Hi,

Sorry for the delayed response.
I have been doing some tests and it's still don't working from me.

Here I'm attaching the code where I create the connection and I disable the ssl verification.

```
BASE_URL_REGION=https://eu-de.iaas.cloud.ibm.com/v1
connector_cloud = VpcV1(authenticator=authenticator)
connector_cloud.set_service_url(BASE_URL_REGION)
print("SET DISABLE SSL VERIFICATION")
connector_cloud.set_disable_ssl_verification(True)
```

Here's the output when I try to do the connection:

```
<bound method BaseService.set_disable_ssl_verification of <ibm_vpc.vpc_v1.VpcV1 object at 0x00000183428AD8E0>>
2
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): iam.cloud.ibm.com:443
send: b'CONNECT iam.cloud.ibm.com:443 HTTP/1.0\r\n'
send: b'\r\n'
{SSLError(MaxRetryError("HTTPSConnectionPool(host='iam.cloud.ibm.com', port=443): Max retries exceeded with url: /identity/token (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1123)')))"))}
```

[...]

PR #187, opened 2024-01-23
fix: use the correct SSL config if cert verification is disabled

pyrooka:
It turned out, disabling the SSL certificate verification didn't work as expected.
We passed the right arguments to the `requests` package, but since we use
a custom SSL context, it couldn't turn off the verification entirely, because the
default SSL context has the `check_hostname` set to `True` and the documentation
says the following: "`verify_mode` is now automatically changed to `CERT_REQUIRED`
when hostname checking is enabled and `verify_mode` is `CERT_NONE`."
This commit adds an addition step which disables the hostname checking in our
custom SSL context.

Ref: https://github.com/IBM/python-sdk-core/issues/186


test_integration/test_no_ssl_verification.py
```diff
@@ -0,0 +1,39 @@
...
+import os
+import threading
+from http.server import HTTPServer, SimpleHTTPRequestHandler
+from ssl import PROTOCOL_TLS_SERVER, SSLContext
+
+from test.test_base_service import AnyServiceV1
+from ibm_cloud_sdk_core.authenticators import NoAuthAuthenticator
+
+
+def test_no_ssl_verification():
```
padamstx, line 11: Should we add a test that uses the same certificate and key, but does NOT disable ssl verification?   I assume this would be a negative test as we should get an exception for that scenario, right?

padamstx (commented):
LGTM, but I have a couple of questions/comments:
1. Perhaps we should add a negative test that shows that an exception occurs (bad certificate) if SSL verification is NOT disabled.
2. Perhaps it would be good to document how you created the certificate and key files used during the test.  Maybe the test_no_ssl_verification.py file would be a good place for this?

test_integration/test_no_ssl_verification.py
pyrooka: Right and it's a good idea. I'll also add the command to generate the cert files.

padamstx (approved):
LGTM

test_integration/test_ssl_verification.py
```diff
@@ -0,0 +1,62 @@
...
+    ssl_context.load_cert_chain(certfile=cert, keyfile=key)
+
+    # Create and start the server on a separate thread.
+    server = HTTPServer(('127.0.0.1', 3333), SimpleHTTPRequestHandler)
+    server.socket = ssl_context.wrap_socket(server.socket, server_side=True)
+    t = threading.Thread(target=server.serve_forever)
+    t.start()
+
+    # We run everything in a big try-except-finally block to make sure we always
+    # shutdown the HTTP server gracefully .
```
pyrooka, line 42:
```suggestion
    # shutdown the HTTP server gracefully.
```

Look what I just found!! @padamstx 😂

padamstx: oh gosh, how did we survive with that extra space? :joy:

PR #189, opened 2024-02-26
fix: adjust IAM token expiration time

padamstx:
This commit changes the IAM, Container and VPC Instance authenticators slightly so that an IAM access token will be viewed as "expired" when the current time is within 10 seconds of the official expiration time. IOW, we'll expire the access token 10 secs earlier than the IAM server-computed expiration time.
We're doing this to avoid a scenario where
an IBM Cloud service receives a request along
with an "almost expired" access token and then uses that token to perform downstream requests in a
somewhat longer-running transaction and then the
access token expires while that transaction is
still active.

### DIFF
--- a/ibm_cloud_sdk_core/token_managers/iam_request_based_token_manager.py
+++ b/ibm_cloud_sdk_core/token_managers/iam_request_based_token_manager.py
@@ -1,6 +1,6 @@
 # coding: utf-8
 
-# Copyright 2019 IBM All Rights Reserved.
+# Copyright 2019, 2024 IBM All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


ibm_cloud_sdk_core/token_managers/iam_request_based_token_manager.py
```diff
@@ -167,3 +168,19 @@ def set_scope(self, value: str) -> None:
             value: A space seperated string that makes up the scope parameter.
         """
         self.scope = value
+
```
padamstx, line 171: This is the key part of the fix.  `_is_token_expired()` is inherited from the TokenManager class.  We'll just override its implementation here to handle the implement the fix for the IAM and Container authenticators.  Specifically, we'll just declare the token expired if the current time is within 10 seconds of the IAM server-reported expiration time.

ibm_cloud_sdk_core/token_managers/vpc_instance_token_manager.py
```diff
@@ -152,3 +153,19 @@ def retrieve_instance_identity_token(self) -> str:
         logger.debug('Returned from VPC \'create_access_token\' operation."')
 
         return response['access_token']
+
```
padamstx, line 156: Override _is_token_expired() for the VPC token manager also.

pyrooka (approved):
Looks good to me!

dpopp07 (approved):
Looks good! 👍

PR #193, opened 2024-04-17
refactor(BaseService): improve prepare_request() docstring

padamstx:
This commit simply modifies the docstring comments related to the `headers` and `params` arguments to the `BaseService.prepare_request()` function in order to clarify that any entries in either of the two dictionary arguments will simply be ignored if they have a value of None.


dpopp07 (approved):
Looks good! 👍

PR #194, opened 2024-06-03
fix: specify the maximum version of `requests` to avoid regression

pyrooka:
Version `2.23.3` of the `requests` package broke our custom `SSLHTTPAdapter`. This commit defines the maximum version of that package, to make sure it will continue working without an issue. Note that, this is a short term solution only. For more details see this issue: https://github.com/psf/requests/issues/6730


padamstx (approved):
LGTM

PR #195, opened 2024-06-19
test: minor updates to a few testcases

padamstx:
This commit makes some minor changes to a few testcases:
1. test_api_exception.py: add tests for some requests.Response operations

2. test_token_manager.py/test_jwt_token_manager.py: modified the abstract base class checks to work in Python 3.12 by matching only part of the error message that appears in the exception

3. I also added python 3.12 to the travis build definition since I found and fixed a couple of unit test errors when using python 3.12

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ dist: jammy
 matrix:
   include:
     - python: 3.8
-    - python: 3.9
     - python: 3.10
     - python: 3.11
+    - python: 3.12
 
 cache: pip3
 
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -5,9 +5,9 @@ dist: jammy
 matrix:
   include:
     - python: 3.8
-    - python: 3.9
     - python: 3.10
     - python: 3.11
+    - python: 3.12
```
padamstx, line 10: It didn't seem right to build on FIVE versions of python so I removed python 3.9 so we would still be building on the current supported min version (3.8).

dpopp07 (approved):
Looks good! 👍

PR #196, opened 2024-07-02
fix: always load the default certs on our custom SSL context

pyrooka:
This is a follow-up PR of [this one](https://github.com/IBM/python-sdk-core/pull/194). In the linked PR we had to specify and older version of the `requests` package to avoid a regression.

This PR contains the fix for that, so we can continue using always the latest version. Furthermore, the HTTPS related tests have been improved and moved to the directory of the unit tests, to always include them in our builds.


ibm_cloud_sdk_core/http_adapter.py
```diff
@@ -0,0 +1,27 @@
...
+    def __init__(self, *args, **kwargs):
+        self._disable_ssl_verification = kwargs.pop('_disable_ssl_verification', None)
+
+        super().__init__(*args, **kwargs)
+
+    def init_poolmanager(self, connections, maxsize, block=DEFAULT_POOLBLOCK, **pool_kwargs):
+        """Create and use custom SSL configuration."""
+
+        ssl_context = create_urllib3_context()
+        ssl_context.load_default_certs()
```
pyrooka, line 20: This is the actual fix, the rest is unchanged.

test/test_http_adapter.py
```diff
@@ -1,47 +1,50 @@
...
-#    -x509 \
-#    -subj "/C=US/CN=localhost" \
-#    -keyout test_ssl.key \
-#    -out test_ssl.cert
+# pylint: disable=line-too-long,pointless-string-statement
+"""
+openssl req -x509 -out test_ssl.crt -keyout test_ssl.key \
+  -newkey rsa:2048 -nodes -sha256 -days 36500 \
+  -subj '/CN=localhost' -extensions EXT -config <( \
+   printf "[dn]\nCN=localhost\n[req]\ndistinguished_name = dn\n[EXT]\nsubjectAltName=DNS:localhost\nkeyUsage=digitalSignature\nextendedKeyUsage=serverAuth")
```
pyrooka, line 22: This command works better for local domain verification. (`localhost`)

test/test_http_adapter.py
```diff
@@ -50,13 +53,29 @@ def test_ssl_verification():
...
         res = service.send(prepped)
         assert res is not None
+
+        # Lastly, try with an external URL.
+        # This test case is mainly here to reproduce the regression
+        # in the `requests` package that was introduced in `2.32.3`.
+        # More details on the issue can be found here: https://github.com/psf/requests/issues/6730
+        service = BaseService(service_url='https://raw.githubusercontent.com', authenticator=NoAuthAuthenticator())
+        assert service.disable_ssl_verification is False
+        prepped = service.prepare_request('GET', url='/IBM/python-sdk-core/main/README.md')
```
pyrooka, line 73: I'm not too happy with this solution, but I found this to be the most reliable way to reproduce the problem. Actually the only way because I was not able to trigger the exception in any of my local experiments. I totally understand if you have concerns about this, so let's discuss.

test/test_http_adapter.py
```diff
@@ -1,62 +1,81 @@
...
-        service = BaseService(service_url='https://127.0.0.1:3333', authenticator=NoAuthAuthenticator())
+        service = BaseService(service_url='https://localhost:3333', authenticator=NoAuthAuthenticator())
         #
         # First call the server with the default configuration.
-        # It should fail due to the invalid SSL cert.
+        # It should fail due to the self-signed SSL cert.
         assert service.disable_ssl_verification is False
         prepped = service.prepare_request('GET', url='/')
-        with pytest.raises(SSLError):
+        with pytest.raises(SSLError, match="certificate verify failed: self-signed certificate"):
```
pyrooka, line 53: Just a small tweak to make sure we catch the exact same error that we want.

test/test_http_adapter.py
```diff
@@ -0,0 +1,107 @@
...
+    res = service.send(prepped)
+    assert res is not None
+
+    # Lastly, try with an external URL.
+    # This test case is mainly here to reproduce the regression
+    # in the `requests` package that was introduced in `2.32.3`.
+    # More details on the issue can be found here: https://github.com/psf/requests/issues/6730
+    service = BaseService(service_url='https://raw.githubusercontent.com', authenticator=NoAuthAuthenticator())
+    assert service.disable_ssl_verification is False
+    prepped = service.prepare_request('GET', url='/IBM/python-sdk-core/main/README.md')
```
ricellis, line 105: >I'm not too happy with this solution, but I found this to be the most reliable way to reproduce the problem. Actually the only way because I was not able to trigger the exception in any of my local experiments. I totally understand if you have concerns about this, so let's discuss.

The only other thing I can think of is listing the certs available and checking it is more than 0, but I don't know if we have enough visibility to be able to do that.

Certainly using a website that has a public certificate chain leading to some default root cert is a surefire way to prove that it can successfully negotiate a connection using the loaded default certs. However, whilst connecting to github proves we have some default certs it doesn't necessarily prove we have the certificate we need so perhaps switching to some `cloud.ibm.com` site would offer more value since it will likely use the same root certifcate chain as our services do? Maybe something like `https://cloud.ibm.com/status` or with a filter like `https://cloud.ibm.com/status/security?component=ibm-cloud&location=global` if it's too large a page otherwise?

ricellis (approved):
I think this is a good change to workaround the requests issues. I supect the comment wasn't directed at me, but I had one thought around the external wesbite test, but I wouldn't consider that a blocker to merging this anyway.

test/test_http_adapter.py
pyrooka: > The only other thing I can think of is listing the certs available and checking it is more than 0, but I don't know if we have enough visibility to be able to do that.

That's a clever idea but I can't tell either - at least off the to of my head. I think I'll take a look because this sounds interesting.

> ...perhaps switching to some `cloud.ibm.com` site would offer more value...

Agreed, it makes sense!

dpopp07 (approved):
All looks good to me! 👍

padamstx (approved):
LGTM, but there are test failures in the Travis build :(

test/test_http_adapter.py
```diff
@@ -0,0 +1,123 @@
...
+    # In some cases (especially in Ubuntu containers that we use for testing on Travis)
+    # the default CA certifications are stored in a different place, so let's try to
+    # load those before making the final decision for this test case.
+    if len(ssl_context.get_ca_certs()) == 0:
+        try:
+            default_ca_path = get_default_verify_paths().capath
+            ssl_context.load_verify_locations(os.path.join(default_ca_path, 'ca-certificates.crt'))
+        except:
+            # Errors are ignored, let's jump straight to the assertion.
+            pass
```
pyrooka, line 117: This is the change that I added to fix the builds. I was thinking about to implement this (or similar) logic in our custom HTTP adapter, but decided not to do it. The reason is, I want to keep our cert handling behavior as close to the original as possible. Let me know if you have any concerns with this!

test/test_http_adapter.py
```diff
@@ -0,0 +1,123 @@
...
+    # This test case is mainly here to reproduce the regression
+    # in the `requests` package that was introduced in `2.32.3`.
+    # More details on the issue can be found here: https://github.com/psf/requests/issues/6730
+    service = BaseService(service_url='https://cloud.ibm.com', authenticator=NoAuthAuthenticator())
+    assert service.disable_ssl_verification is False
+
+    ssl_context = service.http_adapter.poolmanager.connection_pool_kw.get("ssl_context")
+    assert ssl_context is not None
+    # In some cases (especially in Ubuntu containers that we use for testing on Travis)
+    # the default CA certifications are stored in a different place, so let's try to
```
padamstx, line 109: A really minor nit...
```suggestion
    # the default CA certificates are stored in a different place, so let's try to
```

padamstx (approved):
LGTM

PR #197, opened 2024-07-11
fix: improve the detection and loading of default certificates

pyrooka:
Follow up PR of https://github.com/IBM/python-sdk-core/pull/196. Turned out in some cases - especially in containers - the certificate verification is still failing due to the missing certs. It's because the location of those files really depend on the system and the OpenSSL configuration.
This PR adds a workaround for this problem, by loading the default certs from the location that the `certifi` package reports. The `requests` package uses the same logic, so it should cause no issues.

### DIFF
--- a/ibm_cloud_sdk_core/http_adapter.py
+++ b/ibm_cloud_sdk_core/http_adapter.py
@@ -1,10 +1,10 @@
 import ssl
 
+import certifi
 from requests.adapters import HTTPAdapter, DEFAULT_POOLBLOCK
 from urllib3.util.ssl_ import create_urllib3_context
 
 
-# pylint: disable=fixme
 class SSLHTTPAdapter(HTTPAdapter):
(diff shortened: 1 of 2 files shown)


ricellis (commented):
My only comment would be whether `certifi` should be added to the `requirements.txt` since it is now referenced directly, but is included transitively via `requests`. I suppose another option might be using `requests.certs.where()` to avoid the reference to `certifi`. I suppose another advantage of that might be that anyone using a custom `requests` package would get their expected behaviour - see e.g. https://github.com/psf/requests/blob/main/src/requests/certs.py

pyrooka:
That's a really good point! About the requirements, I did not add the `certifi` package on purpose, but now it doesn't matter anymore as we've stopped referencing to that.

ricellis (approved):

padamstx (approved):
LGTM

PR #198, opened 2024-07-12
test: use the correct `ConnectionError` class in the tests

pyrooka:
A recently introduced unit test uses the `ConnectionError` class from the standard library
to check the type of a returned exception, but that's not correct because the actual
exception that can be thrown there is the `requests.exceptions.ConnectionError`.
This confusion made the tests fail sometimes.


padamstx (approved):
LGTM

PR #204, opened 2024-07-29
build: add detect-secrets stage to build

padamstx:


dpopp07 (approved):
Looks good! 👍

.travis.yml
```diff
@@ -2,48 +2,54 @@ language: python
...
+        - npm run semantic-release
+
+    - stage: Publish-Release
+      python: "3.8"
+      name: Publish-To-PyPi
+      before_deploy:
+        - pip install bump-my-version
+      deploy:
+        - provider: pypi
+          setuptools_version: "60.8.2"
```
pyrooka, line 51: Given the recent fun with `setuptools`, I'm thinking about to remove this constraint from the builds. I think I was the one who added these because of an issue-that-I-cannot-recall in the past and now I am wondering if it'd be safe to remove these. `setuptools` released 12 major versions since then.
This is merely a question, this PR has nothing to do with it.

pyrooka (approved):
LGTM!

.travis.yml
padamstx: We explicitly install setuptools in .travis.yml and we also have that setting as part of the deploy step.  Perhaps we should just add setuptools as a dependency to requirements.txt (or requirements-dev.txt), then remove the explicit install command from .travis.yml.  What do you think?

PR #210, opened 2024-09-17
improve message filtering

padamstx:


pyrooka (approved):
Changes are looking pretty good, thanks for making it!

PR #213, opened 2024-10-29
feat(HTTP Client): Migrate Requests to Niquests

Ousret:
This PR effectively replace the http client Requests for [Niquests](https://github.com/jawah/niquests).
Niquests is a drop-in replacement for Requests that is no longer under feature freeze.

This new client support HTTP/2, and HTTP/3 by default and offers both sync and async interfaces. It supports all the latest shiny features you would expect from an http client.

If you were interested on merging, I will be interested to propose a followup PR that will propose a script that generate
the async part of this SDK. If you want to.

Why make the switch?

I believe this will benefit you and your users:

- Lower burden on maintaining library by having both sync/async in a single core package with a single code base.
- Faster HTTP, multiplexed connection, HTTP/3 ready, etc..
- Huge accent on the security, strong cipher enforced, TLS 1.2+, Post-Quantum KX over QUIC, OCSP validation, OS Trust Store, etc...

*disclaimer:* _I maintain Niquests._

### DIFF
--- a/ibm_cloud_sdk_core/api_exception.py
+++ b/ibm_cloud_sdk_core/api_exception.py
@@ -18,7 +18,7 @@ import warnings
 from http import HTTPStatus
 from typing import Optional
 
-from requests import Response
+from niquests import Response
 
 
 class ApiException(Exception):
--- a/ibm_cloud_sdk_core/authenticators/basic_authenticator.py
+++ b/ibm_cloud_sdk_core/authenticators/basic_authenticator.py
@@ -16,7 +16,7 @@
 
 import base64
 
-from requests import Request
+from niquests import Request
 
 from ibm_cloud_sdk_core.logger import get_logger
 from .authenticator import Authenticator
--- a/ibm_cloud_sdk_core/authenticators/bearer_token_authenticator.py
+++ b/ibm_cloud_sdk_core/authenticators/bearer_token_authenticator.py
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from requests import Request
+from niquests import Request
 
 from ibm_cloud_sdk_core.logger import get_logger
 from .authenticator import Authenticator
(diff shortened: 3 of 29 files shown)


pyrooka:
Hi,

First off thanks for the effort and your PR. Although your project looks interesting and you mentioned some good points in the description, we don't have a plan to do this migration right now. Mainly because the current set of features satisfy the requirements of each package that relies on this package. And even though your project is a drop-in-replacement for `requests`, a lot of service SDKs depend on this project so we need to be very careful, especially with such a big change.

I leave this PR open for now, so that other maintainers can take a look too, but I expect it to be closed soon.

Ousret:
Thanks for your fast response.

I knew the chances were slim. But not presenting anything clearly have a worth outcome.

> Mainly because the current set of features satisfy the requirements of each package that relies on this package.

I saw that async is clearly a recurring subject on the various repositories out there, and how you take security seriously. So I though that a opening was possible.

> And even though your project is a drop-in-replacement for requests, a lot of service SDKs depend on this project so we need to be very careful, especially with such a big change.

Not necessarily that complicated. You could have a major version, then downstream could at his time and convenience unpin the major version when they verified that everything is fine.

I can answer and guide you through this challenging change.

Regards,

pyrooka:
> I knew the chances were slim. But not presenting anything clearly have a worth outcome.

Yep, definitely! :)

> I saw that async is clearly a recurring subject on the various repositories out there, and how you take security seriously. So I though that a opening was possible.

It will be, one day. We have other public/private projects that we need to keep sync together with this - in terms of functionality -, so that's why we decided to hold this back for now.

> Not necessarily that complicated. You could have a major version, then downstream could at his time and convenience unpin the major version when they verified that everything is fine.

Yeah, we already use semantic versioning and we are aware of this possibility, but that would mean we need to maintain 2 different versions - at least for a while.

Anyway, I close this PR for now (we may reopen it in the future), but your effort is much appreciated!

PR #233, opened 2026-06-05
feat(auth): add support for new service version in VPC Instance Auth

Andris28:
This PR adds support for the new VPC Instance Metadata Service version (2025-08-26) with the following changes:

- Added configurable `ServiceVersion` field (default: "2022-03-01") that switches to new API paths (`/identity/v1/token`, `/identity/v1/iam_tokens`) when set to "2025-08-26"
- Added `TokenLifetime` field (default: 300s) for customizable token expiration
- Configuration via builder methods (`SetServiceVersion()`, `SetTokenLifetime()`) or `VPC_IMS_VERSION` environment variable
- **Full backward compatibility with older service versions**

### DIFF
--- a/Authentication.md
+++ b/Authentication.md
@@ -473,11 +473,12 @@ authenticator = VPCInstanceAuthenticator(iam_profile_crn='crn:iam-profile-123')
 service = ExampleServiceV1(authenticator=authenticator)
 
 # 'service' can now be used to invoke operations.
+```
 
-# To use the new service version with custom token lifetime.
+To use the new service version with custom token lifetime:
+```python
 authenticator = VPCInstanceAuthenticator(
     iam_profile_id='iam-profile-id-123',
-    url='https://api.metadata.cloud.ibm.com',
     service_version='2025-08-26',
     token_lifetime=900  # 15 minutes
 )
@@ -488,12 +489,15 @@ External configuration:
 ```
 export EXAMPLE_SERVICE_AUTH_TYPE=vpc
 export EXAMPLE_SERVICE_IAM_PROFILE_CRN=crn:iam-profile-123
+```
 
 To use the new service version:
-export EXAMPLE_SERVICE_AUTH_TYPE=vpc
-export EXAMPLE_SERVICE_IAM_PROFILE_ID=iam-profile-id-123
-export EXAMPLE_SERVICE_SERVICE_VERSION=2025-08-26
 ```
+export EXAMPLE_SERVICE_AUTH_TYPE=vpc
+export EXAMPLE_SERVICE_IAM_PROFILE_CRN=crn:iam-profile-123
+export EXAMPLE_SERVICE_SERVICE_VPC_IMS_VERSION=2025-08-26
+```
+
 Application code:
 ```python
 from <sdk-package-name>.example_service_v1 import *
--- a/Authentication.md
+++ b/Authentication.md
@@ -440,7 +440,7 @@ The IAM access token is added to each outbound request in the `Authorization` he
 
 - iam_profile_id: (optional) the id of the linked trusted IAM profile to be used when obtaining the IAM access token.
 
-- url: (optional) The VPC Instance Metadata Service's base URL.
+- url: (optional) The VPC Instance Metadata Service's base URL.  
 The default value of this property is `http://169.254.169.254`. However, if the VPC Instance Metadata Service is configured
 with the HTTP Secure Protocol setting (`https`), then you should configure this property to be `https://api.metadata.cloud.ibm.com`.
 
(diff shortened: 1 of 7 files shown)


Authentication.md
```diff
@@ -440,10 +440,17 @@ The IAM access token is added to each outbound request in the `Authorization` he
 
 - iam_profile_id: (optional) the id of the linked trusted IAM profile to be used when obtaining the IAM access token.
 
-- url: (optional) The VPC Instance Metadata Service's base URL.  
+- url: (optional) The VPC Instance Metadata Service's base URL.
```
pyrooka: The trailing spaces shouldn't be removed. When there is no spaces or multiple newlines characters between two rows, they are merged into a single line.

Authentication.md
```diff
@@ -460,7 +467,12 @@ from ibm_cloud_sdk_core.authenticators import VPCInstanceAuthenticator
 from <sdk-package-name>.example_service_v1 import *
 
 # Create the authenticator.
-authenticator = VPCInstanceAuthenticator(iam_profile_crn='crn:iam-profile-123')
+authenticator = VPCInstanceAuthenticator(
+    iam_profile_id='iam-profile-id-123',
+    url='https://api.metadata.cloud.ibm.com',
+    service_version='2025-08-26',
+    token_lifetime=900  # 15 minutes
+)
```
pyrooka, line 484: A new example that shows the usage of the new service version would be a bit better - instead of updating this one.

ibm_cloud_sdk_core/get_authenticator.py
```diff
@@ -127,6 +127,7 @@ def __construct_authenticator(config: dict) -> Authenticator:
             iam_profile_crn=config.get('IAM_PROFILE_CRN'),
             iam_profile_id=config.get('IAM_PROFILE_ID'),
             url=config.get('AUTH_URL'),
+            service_version=config.get('SERVICE_VERSION'),
```
pyrooka: I think this name is too generic. `VPC_IMS_VERSION` would be clearer and less likely to cause a conflict with an existing environment variable.

ibm_cloud_sdk_core/authenticators/vpc_instance_authenticator.py
```diff
@@ -55,15 +55,32 @@ class VPCInstanceAuthenticator(Authenticator):
...
 
     def __init__(
-        self, iam_profile_crn: Optional[str] = None, iam_profile_id: Optional[str] = None, url: Optional[str] = None
+        self,
+        iam_profile_crn: Optional[str] = None,
+        iam_profile_id: Optional[str] = None,
+        url: Optional[str] = None,
+        *,
+        token_lifetime: Optional[int] = None,
+        service_version: Optional[str] = None,
```
pyrooka, line 67: You can define default values for the arguments, so you don't have to check their emptiness later.

```suggestion
        token_lifetime: Optional[int] = VPCInstanceTokenManager.DEFAULT_TOKEN_LIFETIME,
        service_version: Optional[str] = VPCInstanceTokenManager.DEFAULT_SERVICE_VERSION,
```

ibm_cloud_sdk_core/token_managers/vpc_instance_token_manager.py
```diff
@@ -52,22 +52,42 @@ class VPCInstanceTokenManager(JWTTokenManager):
...
 
     def __init__(
-        self, iam_profile_crn: Optional[str] = None, iam_profile_id: Optional[str] = None, url: Optional[str] = None
+        self,
+        iam_profile_crn: Optional[str] = None,
+        iam_profile_id: Optional[str] = None,
+        url: Optional[str] = None,
+        *,
+        token_lifetime: Optional[int] = None,
+        service_version: Optional[str] = None,
```
pyrooka, line 74: Same as with the token manager.


```suggestion
        token_lifetime: Optional[int] = self.DEFAULT_TOKEN_LIFETIME,
        service_version: Optional[str] = self.DEFAULT_SERVICE_VERSION,
```

pyrooka (changes requested):
Left a few comments to address, but it looks good overall.

ibm_cloud_sdk_core/authenticators/vpc_instance_authenticator.py
Andris28: There is a possibility where in the `get_authenticator.py` we would explicitly pass `None` as a parameter to the `VPCInstanceAuthenticator`. In that case the default parameter value gets overwritten, which would cause the version to be missing. Correct me if I'm wrong but, we would need to check for emptiness anyways.

pyrooka: You're right, thanks for pointing this out. Please dismiss my suggestion.

ibm_cloud_sdk_core/token_managers/vpc_instance_token_manager.py
```diff
@@ -52,22 +52,42 @@ class VPCInstanceTokenManager(JWTTokenManager):
         url (str, optional): The VPC Instance Metadata Service's base endpoint URL.
     """
 
-    METADATA_SERVICE_VERSION = '2022-03-01'
+    DEFAULT_SERVICE_VERSION = '2022-03-01'
```
pyrooka: I think it would be better not to rename this constant. This could potentially break others' code if they use this directly. I know the original name is a bit inconsistent, but it's already there and I'd like to keep it as-is.

ibm_cloud_sdk_core/token_managers/vpc_instance_token_manager.py
```diff
@@ -81,7 +101,10 @@ def request_token(self) -> dict:
         # Retrieve the Instance Identity Token first.
         instance_identity_token = self.retrieve_instance_identity_token()
 
-        url = self.url + '/instance_identity/v1/iam_token'
+        if self.service_version == "2025-08-26":
```
pyrooka: Since this new version is used in multiple places, it might be a good idea to put it into a new constant. (DRY :))

(I'm just about to update my PR, because I realized that I do it similarly. 😅)

pyrooka (changes requested):
Left two small comments that would be nice to discuss/address before merging.

Also, sorry for the late review. I wrote my comments hours ago, but I forgot to send the review itself... 🤦

pyrooka (approved):
LGTM!

PR #234, opened 2026-06-18
build: use dependency cooldown logic

Andris28:
In this commit we add dependency cooldown logic
that enables us to update more frequently and worry
less about compromised packages.


renovate.json
```diff
@@ -0,0 +1,18 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": ["config:recommended"],
+  "prConcurrentLimit": 4,
+  "reviewers": ["Norbert-Biczo", "Lidia-Tarcza", "Andras-Felleg"],
```
diatrcz: I don't think internal gh usernames will work on the external github, but correct me if im wrong.

diatrcz (changes requested):

diatrcz (approved):
lgtm!

pyrooka (approved):
Looks good!

